PR #31, opened 2026-06-08
Fix nixling switch/boot/test: emit closure hostGeneration so store-view intents exist

vicondoa:
## Problem

Daemon-mode `nixling switch` / `boot` / `test` always failed with
`broker-error` (exit 78). The broker audit log shows the real error:

```
op=RunActivation disposition=bundle-intent-missing
error_kind=Broker.BundleIntentMissing
error_message="no store-view intent in the trusted bundle for opaque id `<vm>`"
```

The only way to apply a new VM generation was the
`nixling down <vm> --apply` + `nixling up <vm> --apply` workaround.

## Root cause

`nixos-modules/closures-json.nix` has emitted the per-VM closure
`generation` block as **all-null** (`hostGeneration: null`) since the
v1.0 daemon-only rewrite. The broker resolver's
`build_store_view_intents` **skips** any closure whose
`host_generation` is null, so the trusted bundle ended up with **zero**
store-view intents.

`RunActivation` (switch/boot/test) hard-requires a store-view intent →
always errored. `vmStart` (up/down) only prepares one `if let Some(intent)`,
so it tolerated the absence — which is exactly why down+up worked.

Verified against the live bundle (`/etc/nixling/closures/<vm>.json`) and
the broker audit log.

## Fix

- **`closures-json.nix`** now emits a deterministic, content-derived
  `hostGeneration` = `sha256(toplevel path) % (2^32-1) + 1` (non-zero
  u32). Stable per closure, changes when the closure changes, no runtime
  state. Verified end-to-end (built the `corp-vm` closure →
  `"hostGeneration":3271411828`).
- **Collision hardening.** Because the generation is a u32 content hash,
  two distinct closures of one VM could astronomically-rarely collide.
  The hardlink-farm generation marker now pins the **closure identity**
  (toplevel basename, via `ResolvedStoreViewIntent::closure_identity()`),
  and `build_farm` **fails closed** (`GenerationCollision`) instead of
  unioning two closures into one store view. A **markerless partial**
  generation dir (crashed earlier build) is rebuilt from scratch rather
  than hardlinked over.
- **Rollback fix.** Rollback now resolves its target store-view basename
  from the rolled-back generation's **own marker** instead of the
  current bundle intent, so rolling back to a different closure no longer
  looks for a non-existent `generations/<old-gen>/<current-basename>`.
- Corrected the `StoreSyncRequest` "generations are monotonic" doc
  comment to "content-derived equality token"; regenerated
  `wire-protocol.json` + `daemon-api.md`.
- CHANGELOG `[Unreleased]` Fixed entry.

## Tests

New regression tests: `build_farm_refuses_generation_collision`,
`build_farm_idempotent_for_same_closure`,
`build_farm_rebuilds_markerless_partial_generation`,
`closure_identity_uses_toplevel_basename`,
`rollback_target_view_path_uses_rolled_back_marker_basename`.

Validation: both cargo workspaces build; `cargo test --workspace` +
broker `cargo test` (default + `layer1-bootstrap`) green; clippy clean
on changed files; Nix parse + closure emitter build OK; schema-drift
gate green.

## Out of scope (pre-existing, not touched)

These exist on `main` independently of this change:
- broker clippy errors at `runtime.rs:3994` and `disk_init.rs:496-498`
- CLI-schema drift in `docs/reference/cli-output/status.schema.json`
  (`gen-cli-schemas` wants `ApiReadyStatusV1` added)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,21 @@ deprecations ship one minor release before removal.
 
 ## [Unreleased]
 
+### Fixed
+
+- `nixling switch` / `boot` / `test` no longer fail with `broker-error`
+  ("no store-view intent in the trusted bundle"). The per-VM closure
+  artifact now emits a populated `hostGeneration` (a deterministic,
+  content-derived store-view generation), so the broker builds a
+  store-view intent for every VM instead of skipping it. Previously
+  live activation was impossible and the only way to apply a new
+  generation was `nixling down <vm> --apply` followed by
+  `nixling up <vm> --apply`. The per-VM `/nix/store` hardlink farm now
+  also fails closed on a store-view generation collision (two distinct
+  closures of one VM mapping to the same generation number) instead of
+  unioning them, by pinning the closure identity in the generation
+  marker.
+
 ## [1.2] - 2026-06-03
 
 Primarily a stabilization release per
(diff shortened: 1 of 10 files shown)


vicondoa:
Merging via admin override. The three red checks (`tier0`/ci-coverage orphaned `tests/privileges-json-rust-vs-nix-eval.sh` from the v1.2 release commit, `eval-shell-tests`/broker-caps-eval, and `eval`/with-entra-id unlocked lock-file input) are **pre-existing on main and unrelated to this diff** — none of those files are touched here. The Rust workspaces, full `cargo test --workspace` + broker tests (default + layer1-bootstrap), clippy on changed files, and the Nix closure emitter were all validated locally.

PR #37, opened 2026-06-09
virtiofsd: allow broker-prepared vms runtime dir

vicondoa:
## Summary

- Align virtiofsd minijail writable runtime dir with the broker-prepared socket root under `/run/nixling/vms/<vm>`.
- Fixes `SpawnRunner failed at virtiofsd-nl-hkeys` when the broker prepares per-VM runtime sockets under the vms runtime root.

## Validation

- `nix eval --impure --raw --expr 'let f = builtins.getFlake (toString ./.); in if f ? nixosModules then "ok" else throw "missing modules"'`


vicondoa:
Closing: the host now uses GitHub nixling main and work-aad starts without this diagnostic branch. This was created during entrablau integration troubleshooting and is not needed for the final rollout.

PR #63, opened 2026-06-17
observability: ADR 0033 follow-ups — host OTLP ingest namespace fix + deployment.environment scheme

vicondoa:
## Summary

Fixes a `226/NAMESPACE` activation failure introduced by ADR 0033's host
OTLP ingest path.

When `nixling.observability.host.otlpIngest.enable = true`, the collector
unit sets `ReadWritePaths = [ "/run/nixling/otel/ingest" ]`, but the ingest
dir was only created in the collector's `ExecStartPre` — which runs *inside*
the unit's mount namespace. systemd builds the `ReadWritePaths` bind mount at
namespace-construction time, **before any `ExecStartPre`**, so the missing
dir fails the unit:

```
nixling-host-otel-collector.service: Failed to set up mount namespacing:
/run/nixling/otel/ingest: No such file or directory
Failed at step NAMESPACE ... status=226/NAMESPACE
```

### Fix

Create `/run/nixling/otel/ingest` via `systemd.tmpfiles.rules` (gated on
`otlpIngest`) so it exists before the unit's namespace is built; `runtimePrep`
still refines its perms/ACLs and unlinks a stale socket. `host-egress.sock`
isolation is unchanged.

### Validation

- Reproduced and confirmed the fix live on a host: with the ingest dir
  pre-created the collector reaches "Everything is ready. Begin running and
  processing data." with `prometheus` + `journald` + `filelog` receivers up.
- New regression coverage: eval cases assert `tmpfilesHasIngest`
  (true with `otlpIngest`, false by default); `policy_state.rs`
  `host_otlp_ingest_socket_isolation` gains a source `want` for the tmpfiles
  rule. `cargo test -p nixling-contract-tests --test policy_state` passes (6).

Hotfix for an activation regression in already-merged ADR 0033 work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,14 @@ deprecations ship one minor release before removal.
 
 ### Changed
 
+- `deployment.environment` is now machine-and-env aware: the central
+  collector stamps it `<hostName>` for host telemetry and
+  `<hostName>-<env>` for workload VMs (e.g. `ddbus`, `ddbus-work`,
+  `ddbus-personal`), instead of the bare host name for everything.
+  `host.name` remains the per-source name (the host's name for host
+  telemetry, the VM's name for workloads). See
+  [ADR 0033](docs/adr/0033-host-collector-parity.md).
+
 - Host-origin telemetry now carries the **hostname** as `vm.name` /
   `host.name` (via `nixling.observability.host.identityName`, default
   `networking.hostName`), assigned at the trusted ingress boundary, rather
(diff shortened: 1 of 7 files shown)


vicondoa:
Added a second commit: `deployment.environment` is now `<hostName>` for host telemetry and `<hostName>-<env>` for workload VMs (e.g. `ddbus`, `ddbus-work`, `ddbus-personal`); `host.name` stays the per-source name (host's name / VM's name). Covered by a new `policy_observability` source gate + reference/CHANGELOG updates.

PR #395, opened 2026-08-08
ci: measure filtered-metadata ARM baselines

vicondoa:
Disposable native ARM measurement after the final offline metadata filter. The job emits normalized artifact baseline fields without store paths and will be closed without merge.


vicondoa:
Final filtered-metadata ARM measurements were recorded successfully; closing this disposable PR without merge.

PR #396, opened 2026-08-08
ci: measure dependency-free walker ARM baselines

vicondoa:
Disposable native ARM measurement after the final dependency-free walker metadata change. It will be closed without merge after recording normalized baseline fields.


vicondoa:
Final dependency-free walker ARM measurements were recorded successfully; closing this disposable PR without merge.

PR #398, opened 2026-08-08
ci: measure Fenix ARM artifact baselines

vicondoa:
Disposable native ARM artifact measurement after pinning Rust 1.97 and closing panel findings. It includes binary content digests and will be closed without merge.


vicondoa:
Reopening as a fresh disposable PR so GitHub schedules the extended ARM measurement workflow.

PR #402, opened 2026-08-08
ci: measure final panel-fix ARM artifacts

vicondoa:
Disposable trusted ARM measurement for the final panel-fix source tip. It emits binary, linkage, closure, and policy identities and will be closed without merge.


vicondoa:
Final panel-fix ARM artifact measurements were recorded successfully; closing this disposable PR without merge.

PR #406, opened 2026-08-09
test: measure Spec 003 ARM artifact closures

vicondoa:
Temporary digest-only native ARM measurement branch for Spec 003 Wave 0. This PR will be closed after collecting the two closure identities.


vicondoa:
Native ARM closure measurements captured; temporary branch is no longer needed.

Issue #530, opened 2026-09-14
host-integration: the lane requires a live sudo timestamp and fails silently in non-interactive runs


vicondoa:
## Context

`make test-host-integration` stages the host-tool bundle with Bazel (cached, fast) and then builds each `vmChecks.<system>.<name>` derivation as `sudo -A -E nix build --option build-users-group "" --option extra-sandbox-paths /dev/vhost-vsock ...`. The `-A` asks for the password through `SUDO_ASKPASS`, which on NixOS is `/run/current-system/sw/bin/systemd-ask-password`.

## Problem

In a non-interactive context - an agent shell, CI, or a session with no ask-password agent - nothing answers that prompt, so **every** check fails after the 90-second timer with:

```
Failed to query password: Timer expired
sudo: no password was provided
sudo: a password is required
```

This happens *before any build starts*, so it reads as a test failure rather than a missing credential, and a full run burns roughly seventeen minutes producing nothing. Worse, `sudo` timestamps are global with `timestamp_timeout=360`, so a run can start fine and then lose authorization mid-flight if one check takes longer than six minutes, turning the later checks into the same confusing failure.

## Acceptance

- A non-interactive invocation reports the credential problem once, immediately, instead of as eleven per-check failures - a preflight `sudo -n true` with an explicit message is enough.
- The target documents that it needs a live sudo timestamp, since one authentication covers the serially-run checks while no single check exceeds the timeout.

Issue #500, opened 2026-09-11
Remove migration scaffolding: one execution pattern, no adapters kept for the rewrite's convenience


vicondoa:
# Remove the migration scaffolding: one execution pattern, no adapters kept for the rewrite's convenience

## Summary

This rewrite was staged so each step could land independently. That staging left
a layer of adapters, compatibility seams, placeholder port implementations, dead
code held open by `#[allow(dead_code)]`, and one-off test doubles. Some are
legitimately still in use by unconverted paths; most exist only because the
migration passed through them. The target state is one execution model, one
effect-port pattern, one shared test double per port shape - no abstraction that
exists to make the migration easy rather than to make the code correct.

## What is there now (evidence, 2026-09-10)

**Adapter modules named after the migration step**

- `packages/d2bd/src/resource_runtime/volume_effect_adapter.rs` (+ its test
  `packages/d2bd/tests/volume_effect_adapter.rs`) and a second
  `packages/d2b-host/src/volume_effect_adapter.rs`.
- `packages/d2b-core/src/device_usbip_adapter.rs` (+ test), `packages/d2b-resource-api/src/adapter.rs`,
  `packages/d2b-provider/src/share_adapter.rs`, `packages/d2b-provider-supervisor/src/adapter.rs`
  (+ `tests/production_adapter.rs`), `packages/d2b-session-unix/src/adapter.rs`.
- The rewiring adapters themselves: `packages/d2b-resource-api/src/manager_backend.rs`
  (the Resource API re-pointed at the per-Zone manager) beside the redb path it
  replaced (`store.rs`, `registered.rs`), i.e. two `ResourceStoreBackend`
  implementations alive at once.

**Placeholder port implementations ("Unavailable..." family)**

- `packages/d2bd/src/resource_runtime/interaction_composition.rs:2747` `UnavailableProcessEffectPort`.
- `packages/d2bd/src/resource_runtime/shared_provider_runtime.rs:1043` `UnavailableSharedProviderEffects`
  (also constructed in production at `resource_runtime.rs:3790`, i.e. a
  placeholder on a live path).
- `packages/d2b-resource-api/src/service.rs:139` `UnavailableUpgradeDispatcher`.
- `packages/d2bd/src/provider_effects.rs:1568` `UnavailableReconciliationEffect` (test-side).

**Dead code kept open instead of deleted**

- `packages/d2bd/src/composition.rs` carries a graveyard of
[...]

PR #541, opened 2026-09-18
refactor: remove the residual migration scaffolding

vicondoa:
## Summary

Removes the migration scaffolding that the v3 resource-runtime rewrite left behind, completing the remaining acceptance criteria of issue #500 on top of what PR #537 already deleted. Plan: `docs/plans/2026-09-18-001-refactor-residual-scaffolding-removal-plan.md`.

Four shapes of work:

1. **One shared recording double per effect-port shape.** The guest, wayland-policy, host, user, provider, and notification-provider crates move their inline driver-effect test doubles into a `src/test_support.rs` module behind a `test-support` cargo feature, with a `*_test_support` Bazel target carrying `crate_features = ["test-support"]`, consumed by both their own tests and the daemon's plane tests. No crate keeps a second implementation of the same port, including the registration-boundary tests that were building descriptors from local stubs.
2. **Every surviving `#[allow(dead_code)]` states a factual reason at its site.** Notes that pointed at a possible future consumer or at a plan unit that has since landed were replaced with the actual state; the module-scope allowance in the shared test harness carries its reason; the generated ttrpc allowance is recorded as a generated-code exception because the generator emits it.
3. **Zero-caller public API deleted.** The per-env USBIP autostart seam in `d2bd-runtime` (667 lines, eleven tests) with its runtime-boundary file-table row and its two ADR dossier rows; the resource runtime's internal routing request enum and channel-backed endpoint wrapper (five in-module tests reworked onto a local stub with their assertions intact); the resource compiler's `compile_provider_artifact` alias; the controller's `BindingChildReconciler` (four tests and one helper whose only subject it was, plus a restored focused test for the child-resource rejection paths that survive). One further unreachable error variant went with the reconciler.
4. **The dead-code scan is runnable.** `cargo-shear` is provisioned in the development shell and in the Bazel compatibility shell the Make dispatcher re-enters, and `check-dead-code` is a classified local goal, so the target finds the scanner instead of failing on a missing tool. `cargo-hawk` stays outside the shell: it needs a nightly toolchain built against rustc internals, which the pinned stable toolchain cannot provide.

Knowingly retained surfaces are recorded with their reasons in `changelog.d/residual-scaffolding-removal.md`: the neutral volume effect-port contract and its host wrapper (a scheduled interface owned by unit `ADR046-vl-012` / task T469), `WatchSink`, `LiveControllerSessionEvidence`, the store's audit-log history read path, and the resource-runtime identity re-export module.

## Testing

- `make check` (Layer-1 Bazel facade, `--config=local`): **923 tests pass** on the head.
- Focused suites during implementation: `cargo test -p d2b-provider-guest` (17 driver + 5 registration), `-p d2b-provider-host|user|provider` (unit + registration each), `-p d2b-resource-runtime` (131), `-p d2b-resource-compiler` (20), `-p d2b-core-controller` (4 binding-child tests), `cargo test -p d2bd-runtime` (462 lib + 2 boundary) - all green.
- Generated inputs regenerated for the new dependency edges: `cargo xtask gen-package-policy-inputs --write`, then `--check` clean.
- `make test-host-integration` (run from a session holding the live sudo timestamp the lane requires): **all eleven vmChecks PASS, lane exit 0** - `bridge-isolation`, `daemon-smoke`, `device-worker-launch`, `guest-agent-cap-confinement`, `guest-shell-service`, `privilege-oracle`, `resource-operator-activation`, `runtime-cloud-hypervisor-guest-preflight`, `state-posture-contract`, `virtiofsd-volume-runtime`, `wayland-proxy`. The run reused the VM closures realized by the immediately preceding attempt, so its per-check wall times (9-19s) are warm-cache figures. (A first attempt in this session failed every check at 90s with `sudo: a password is required` - the lane's documented precondition, tracked as issue #530, not a code failure.)
- Independent review ran in separate contexts (correctness, testing, maintainability, adversarial); every actionable finding it raised is applied in this head, including the two P2s: the dispatcher path that never reached the shell provisioning the scanner, and the deleted test whose subject survived.

## Known Residuals

- **`make check-dead-code` is red by design on this tree, and reads as a delta.** `cargo-shear` reports 81 pre-existing findings repo-wide and `cargo-hawk` cannot run outside a nightly `rustc_private` toolchain; both are identical on the untouched `v3` baseline (verified by running the same scan there). The new baseline note in `docs/contributing/gates-and-lints.md` records this so a future run can distinguish a new finding from the standing corpus. The rustc `dead_code` pass is clean.
- **Three integration test binaries need the feature under cargo.** `d2b-provider-wayland-policy`'s `tests/engine.rs` and `tests/registration.rs`, and `d2b-provider-guest`'s `tests/registration.rs`, declare `required-features = ["test-support"]`, so plain `cargo test -p <crate>` skips them silently. Bazel compiles them through the `*_test_support` targets and runs them; the crate comments and the gates reference now say to use `--features test-support` on the cargo path.

No additional operational monitoring required: the change deletes unreachable public items, moves test-only doubles into feature-gated modules, and provisions a local scanner in the development shell. No daemon, broker, provider runtime, Nix module, generated contract, CLI surface or audit record changes behaviour, and no production dependency closure gains an edge that was not already present (the three promoted dependency edges resolve to nodes already in the production closure, per the regenerated policy inputs).


vicondoa:
Lane evidence on the head (`f5ebcf594`): `make test-host-integration` exits 0 with all eleven vmChecks PASS. An earlier attempt in the same session failed every check at 90s with `sudo: a password is required` - the lane's documented precondition (issue #530) - and the rerun from a session holding a live sudo timestamp is green.

Also on this run: CI's first `rust-broker` result was a flake - `//packages/d2b-broker:d2b_broker_test` failed once on the runner and passed on rerun, and it passes locally across repeats. The aggregate `check` job is now green.

Issue #531, opened 2026-09-14
resource plane: a family that commits a Volume child row must bridge the anchor-refresh hook


vicondoa:
## Context

The plane re-registers per-resource Volume anchors only when a family bridges the `volume_anchor_refresh` hook (`d2b-provider-toolkit/src/shared_provider.rs`): the trait method defaults to `None`, and the child surface refreshes anchors only when the family answers `Some(...)`. A family whose effects commit a Volume or VolumeBinding child row through that surface therefore has to implement it - otherwise the row stays invisible to the plane's anchor cache, and the volume root resolves forever as `volume-anchor` / `source-unresolved`.

Only the network family implemented it. The Device family's TPM state Volume hit exactly that loop: `driver reconcile error kind=volume-layout-effect-failed note=source-unresolved`, `shared_provider_effects: TPM device controller reconcile failed error=Effect(EffectRejected)`, and both `device-worker-launch` and `runtime-cloud-hypervisor-guest-preflight` failed after their wait budgets. Fixed for Device in `7dfe2c083`.

## Why this needs an issue

The mistake is silent where it is made: the family compiles, the effect commits the row, and only a multi-minute VM check reveals it. The trait's documentation states the requirement; nothing enforces it.

## Acceptance

- A test or check fails when a family that commits a Volume/VolumeBinding child row answers `None` from `volume_anchor_refresh`.
- The remaining families (security-key, usbip) are confirmed either to commit no Volume child row or to bridge the hook, and their answer is recorded.

vicondoa:
Closed by #543 (squashed as the merge commit on `v3`).

**What changed:** the plane's anchor projection is driven by the manager's durable-change stream, and the three writer-side refresh obligations are retired — no provider family supplies an anchor-refresh hook, so none can omit one. What the retirement needed on the other side was a failure contract the runtime owns:

- A `NotYet` verdict is a deferral and may name its own retry delay (`DriverFailure::with_retry_after`); a deferral never escalates.
- A retryable operational **error** retries on the actor's ladder, doubling from the configured backoff to a 30 s ceiling and resetting on a converged pass.

**Acceptance evidence:** `make check` 923/923, and the full host integration lane green — all eleven checks, `runtime-cloud-hypervisor-guest-preflight` included. The four parent-row drivers that could not defer without deadlocking their own children keep answering for their convergence, with the not-ready half in the fenced status projection; that constraint, and the cycle it prevents, is recorded in the code.

[...]

PR #543, opened 2026-09-19
refactor(issue-531): drive the anchor projection from the manager change stream

vicondoa:
## Summary

The plane's per-resource Volume anchor projection is a read-side cache that the volume root resolver reads synchronously, so a `Volume` or `VolumeBinding` row committed after the plane's last durable load stayed invisible and its root resolved as `volume-anchor` / `source-unresolved` indefinitely. Three write-side refresh call sites papered over that, one of them a per-provider-family trait hook that a family could forget - the failure that produced issue #531, and that only a multi-minute VM check revealed.

The manager already publishes a durable-change notice from its commit handlers, so the projection now consumes that stream: one subscription selects `Volume` and `VolumeBinding` rows, coalesces drained notices into one bounded per-row re-registration, and relists plus re-registers when the hub reports the cursor is unservable. The three write-side obligations are deleted - the per-family hook chain (trait, re-export, child-surface refresh branch, both family adapters, and the daemon methods they funneled into), the controller child-mutation bridge's per-call registry reload, and the bundle ingest's trailing reload - together with the two structural-ratchet exemptions that named the bridge's retired match arms. An anchor derived from a newer durable row now replaces an earlier one, so an `Updated` Volume whose attachment moved heals instead of keeping the previous guest's identity.

No provider family is asked to remember anything, so none can omit it.

## Validation evidence

- [x] **Exact `make check`:** result=`passed`; evidence=`923/923 tests pass, run per unit commit (891330348, c4cc37430, 19d20ee0e)`.
- [x] **Focused tests for the changed components** were run; list exact commands and results.
      `cargo test -p d2bd --lib` -> 472 passed, 0 failed. Twelve new tests in `packages/d2bd/src/resource_plane_v3.rs`: ten cover the subscription machinery (notice coalescing, the startup handoff, the missed-data and expired-cursor recovery paths, source filtering, sustained traffic, restart), two are route tests asserting that a committed volume's anchor resolves with no writer-side refresh (API apply, controller bridge), plus one registry regression test for the anchor merge.
- [x] **Wider lanes are conditional.** `make test-host-integration` was run on every unit commit with `D2B_VM_CHECK=device-worker-launch` (the check the original incident failed): `PASS 53s` / `PASS 54s`. `make test-integration` was not run - the change touches no container behavior.
- [x] **Changed tests are owner-local:** the new tests are Layer-1 type 2 in the owning crate's `src` test module; no new `tests/*.sh`, gate, or CI job was added.
- [x] **Changelog updated** via `changelog.d/refactor-issue-531-anchor-projection-stream.md`.
- [x] **Docs + CI updated in lockstep:** not applicable - no manifest, schema, generated artifact, workflow, or agent-instruction file changes.

## Notes

- Closes issue #531 under a re-scope its owner accepted: criterion 1 (a test or check fails when a family commits such a row while answering nothing) is satisfied by removing the obligation rather than policing it, and criterion 2 (audit the security-key and usbip families and record their answer) is moot, because a manager-mediated commit publishes regardless of which writer made it.
- Residual, recorded rather than hidden: the hub's selector matches on the resource key alone, so status notices for these rows share the subscription's delivery buffer and a status burst can force repeated full relists (correct, but unbounded in cost; a source-filtered registration would fix it). The refresh is now asynchronous, so the first post-commit resolve retries instead of seeing the anchor immediately. The subscription has no production supervisor yet; a restart is a fresh plane spawning a fresh subscription.
- Deferred: an ordering rule driven by each driver's declared consumed and generated resource types - `reads` and `creations` exist on `DriverDescriptor` but nothing consumes them for ordering, and `creations` under-reports effect-committed rows, so the declaration needed for a sound rule does not exist yet.


vicondoa:
Fresh review of head `19d20ee0e` (required after the fixes applied post-first-review) returned three advisory findings. Recording them here rather than merging over them, and noting where I disagree.

**1. A fresh Volume's first reconcile races the drain window (P2, validator: adversarial).**
Previously the child-surface and bridge refreshes awaited `reload_registry()` in-call, so an anchor existed before the effect returned; the subscription instead needs up to `ANCHOR_DRAIN_WINDOW` (50 ms) before it registers anything. A just-committed Volume's first `resolve_root` can therefore miss and fail `source-unresolved(volume-anchor)`, then succeed on requeue. The suggested fix (make `lookup_anchor` fall through to the store like the socket-target lookups) is **not directly applicable**: `lookup_anchor` is synchronous because the volume root resolver is, while the socket-target fall-through is async. The applicable mitigation is to shorten the lone-commit latency - act on a short quiet window instead of a fixed 50 ms wait - which preserves burst coalescing while removing the fixed 50 ms tax on every commit. Deferred: worth a follow-up, since it changes the coalescing semantics the burst test pins.

**2. The subscription task is detached and never stopped (P3, confidence 100).**
`prepare` spawns it and drops the `JoinHandle`, so dropping the plane neither stops the task nor releases its `Arc` clones of the hub, store, and registry. Re-preparing a zone leaves the previous subscription alive, and both would drain the hub. This is a direct consequence of my removing the handle home: shipping it needed three `allow(dead_code)` on production surface, which this repo's standards forbid. The clean storage-free fix is for the task to hold `Weak` handles and exit when the plane's strong references drop. Deferred to the follow-up that owns supervision.

**3. Sustained ring churn can pin the subscription in a reload loop (P3).**
The cursor advances only on delivered Volume/VolumeBinding changes, while the hub ring evicts on *every* publish, including status traffic for unrelated types. After a `Missed`, the relist cursor can already be evicted, so `register` returns `Expired` again, and under sustained status traffic the loop can stay in reload cycles. Deferred with the hub-side source-filter item below: a registration that can exclude non-durable sources fixes this and the buffer-occupancy cost together.

**Already recorded residuals (unchanged):** the hub's key-only selector shares the delivery buffer with status traffic; the refresh is now asynchronous; no production supervisor yet; two of five commit routes lack hermetic tests; the plan's claim that its tests prove the no-full-sweep bound is overstated.

None of these is a correctness regression against the pre-PR behaviour on the routes the PR exists to fix, and all are either deferred deliberately or blocked on a hub API change. Validation evidence is in the PR body.

vicondoa:
**Not mergeable as it stands.** The full host lane fails on this head, and a baseline run isolates the cause to this change.

Full `make test-host-integration` (all eleven checks), same host, run back to back:

| check | `v3` (72f2be081) | this head (2431ab5e7) |
|---|---|---|
| bridge-isolation | PASS 10s | PASS 8s |
| daemon-smoke | PASS 9s | PASS 48s |
| device-worker-launch | PASS 10s | **FAIL 202s** |
| guest-agent-cap-confinement | PASS 8s | PASS 8s |
| guest-shell-service | PASS 9s | PASS 33s |
| privilege-oracle | PASS 9s | PASS 34s |
| resource-operator-activation | PASS 10s | PASS 73s |
| runtime-cloud-hypervisor-guest-preflight | PASS 19s | **FAIL 346s** |
| state-posture-contract | PASS 18s | PASS 232s |
| virtiofsd-volume-runtime | PASS 10s | PASS 41s |
| wayland-proxy | PASS 9s | PASS 38s |

The baseline is 11/11 green with every check between 8 and 19 seconds. On this head every check is 3-20x slower and the two that fail are exactly the two routes the original incident went through. `device-worker-launch` also passes when run alone on this head, but takes 53s instead of 10s.

**Working diagnosis.** The three retired refresh paths awaited the registry reload in-call, so an anchor existed before the committing effect returned. The subscription instead needs up to `ANCHOR_DRAIN_WINDOW` (50ms) before it registers anything, so a fresh Volume's first `resolve_root` can miss, publish a Failed/Pending status flap, and requeue (~200ms) before it is realized - which inflates every volume-touching check and, under VM load where the drain is slower still, exhausts the bounded waits in these two. This is the same behaviour the adversarial re-review flagged (finding 1) and that I deferred as a residual; the host lane says it is not deferrable.

**Intended fix, not yet applied.** Act on the first notice after a short quiet window (a few milliseconds) instead of a fixed 50ms wait, so a lone commit's anchor lands in roughly the time it used to take synchronously while a tight burst still coalesces into one re-materialization; then re-run the Layer-1 gate and the full host lane on the new head. The alternative - giving the synchronous anchor lookup a store fall-through like the socket-target lookups - is not available without making the resolver async.

Do not merge until the full lane is green on the head.

vicondoa:
Follow-up after the drain-window fix (`f73c2881f`, full lane again on the same host):

| check | `v3` baseline | `2431ab5e7` | `f73c2881f` |
|---|---|---|---|
| device-worker-launch | PASS 10s | FAIL 202s | **PASS 46s** |
| runtime-cloud-hypervisor-guest-preflight | PASS 19s | FAIL 346s | **FAIL 331s** |
| daemon-smoke | PASS 9s | PASS 48s | PASS 47s |
| state-posture-contract | PASS 18s | PASS 232s | PASS 221s |
| resource-operator-activation | PASS 10s | PASS 73s | PASS 73s |
| virtiofsd-volume-runtime | PASS 10s | PASS 41s | PASS 43s |
| wayland-proxy | PASS 9s | PASS 38s | PASS 40s |

Shortening the drain window recovered `device-worker-launch`, so the commit-to-visible race was real and is fixed. It did **not** recover `runtime-cloud-hypervisor-guest-preflight`, and every check is still 2-5x the baseline, which means a second cause is still live. Two candidates, both inspectable from the evidence already on this PR:

1. **Status traffic occupies the subscription's delivery buffer.** The selector matches on the resource key alone, so every `RuntimeStatus` publish for a Volume or VolumeBinding row is queued and filtered only after delivery. On a busy daemon that can raise `Missed`, and a `Missed` costs a full type-scoped reload plus a fresh registration - repeatedly. This is the cascading failure the adversarial re-review described, and it would inflate exactly the busiest checks.
2. **The retired bridge reload also warmed the socket-target cache.** `reload_registry` re-registered VolumeBinding-derived socket targets as well as anchors. Socket lookups do fall through to a bounded store read on a miss, but they no longer get the free warm from every Volume/VolumeBinding commit, so a binding-heavy check can now pay that read many times.

Both point at the same next diagnostic: count how often the subscription relists, and how often a socket-target lookup misses, during `runtime-cloud-hypervisor-guest-preflight` alone. The first fix, if confirmed, is to let the hub filter by change source so non-durable notices never reach this subscription; the second is to have the drain's VolumeBinding path register the derived socket targets it already derives.

Not merging until the full lane is green on the head.

vicondoa:
**Root cause found, and it was this change's own design assumption.** The retirement of the writer-side refreshes assumed a resolution that follows a commit can tolerate the projection's delivery latency. It cannot: the volume root resolver reads the anchor cache **synchronously**, so a miss is a failed layout effect, not a retry.

Instructing `resolve_root` to log the uid it missed made it unambiguous in `runtime-cloud-hypervisor-guest-preflight`:

```
21:24:46.077732Z WARN TEMP anchor miss: the requested volume uid holds no anchor  uid=9fe7ff2b-…   (Volume/state)
21:24:46.078226Z WARN driver reconcile error kind=volume-layout-effect-failed
21:24:46.080961Z INFO TEMP anchor registered                                                  uid=9fe7ff2b-…   name=state
```

**3.2 ms late** — the drain window plus the notice round-trip — and the same pattern repeated thirty seconds later for `acceptance-guest-system`. Downstream, the failed volume layout takes the binding's worker with it: the broker's own spawn invocation times out (`kernel transport: reply timeout`, **7 occurrences against 0 in a passing run**), the worker stays `Pending`, and whichever wait trips first ends the run (`binding-worker-ready` t=168.4 s on one head, `volume-controller-processes` t=203.1 s on another).

The tell that this is an ordering law and not a coverage gap: **a passing log of the same check carries the same three anchor misses and recovers** — because its commit paths refreshed the projection before the effect returned.

**Fix.** The synchronous refresh is restored on the three commit paths the retired reloads covered — the bundle ingest, the controller bridge's child mutations, and the provider child surface (`SharedProviderFamily::volume_anchor_refresh` → `ContextChildSurface::ensure`, the path that creates the device's TPM state volume). What is *not* restored is the cost issue #531 was about: every refresh is now bounded to `Volume` and `VolumeBinding` rows instead of the all-rows sweep, and the subscription stays as the net for commits those paths do not drive, with its recovery reload made zone-open so it matches the coverage of the load it stands in for.

**Tests.** `a_bridge_committed_volume_resolves_with_its_commit` now asserts the anchor is present with **no wait at all** (it fails on the previous head, passes here); `an_api_applied_volume_resolves_through_the_projection` states the net's contract, which is still eventual and still needs its wait. The toolkit's `volume_anchor_refresh_covers_created_and_updated_volume_children` came back with the hook. The plan records the correction under **Implementation Correction (2026-09-19)**, and R7/R8 are marked withdrawn with the evidence that withdrew them.

`make check` is 923/923 on this head; the full host lane is running now.

vicondoa:
**Course corrected on your call — and it is the better answer.**

The approach in the two comments above (restore a writer-side refresh so a resolution cannot miss) is reverted. The projection stands as the only law, and what made that safe before it wasn't is that **a driver failure is now a failure the runtime retries**, with the two laws kept deliberately separate:

- **A `NotYet` verdict is a deferral, not a failure.** The driver may name how long it wants to wait — `DriverFailure::with_retry_after`, carried as `DriverVerdict::NotYet { retry_after }` — and the actor schedules at *that* delay. A deferral that names none keeps the configured backoff, and a deferral never escalates: it is the driver saying "ask me again", not a fault.
- **A retryable operational *error* escalates.** The actor's ladder doubles the wait per consecutive error, from the configured backoff to `DEFAULT_REQUEUE_BACKOFF_MAX` (30 s), and a pass that converges resets it. So a transient error is retried within milliseconds while a persistent one cannot spin.

That is what the host lane was missing. The evidence in the previous comment stands and now reads differently: the anchor landed **3.2 ms** after the resolution, and the run died not on that miss but on the retries behind it — 27 launch failures against a passing run's 11, with the broker's own spawn invocation timing out (`kernel transport: reply timeout`, 7 against 0). A fixed 200 ms retry hammered a transient storm; the ladder rides it out.

Pinned by three tests in `d2b-resource-runtime`:

- `the_retry_ladder_doubles_and_stops_at_its_ceiling` — the ladder's shape, including the clamp.
- `a_deferral_keeps_its_cadence_while_an_error_escalates` — the same 50 ms base delivers ~20 passes for a deferral and a handful for an error, which is the whole point of separating the two.
- `a_driver_named_retry_delay_paces_the_deferral` — a named 400 ms delay governs a 50 ms actor, and the delay rides the structured verdict into the published status.

R7 and R8 stand unchanged. `make check` is 923/923 on this head; the full host lane is running.

vicondoa:
**The provider sweep you asked for, done — and it found more than the one case.**

The rule I enforced, stated once: a reconcile pass that has not realized its row must not claim completion. `Satisfied` is "the desired state is realized, the actor may go ready" — the runtime publishes `Ready` and fires every `WatchCondition::Ready` watcher on that answer. A `Pending`/`Degraded`/`Unknown` row that answers it wakes every dependent while nothing backs the claim, and if the pass also schedules no retry, the row is stranded until an unrelated trigger happens to arrive.

Six sibling investigators audited every provider driver against that rule. What they found, and what now holds:

| crate | was | now |
|---|---|---|
| `d2b-provider-guest` (the CH case) | `Satisfied` with a `Pending` phase; a 30 s resync behind it | `RetryScheduled` + a 2 s **dependency** re-check |
| `d2b-provider-toolkit` (most providers inherit it) | `Satisfied` on any `Pending` family effect | `RetryScheduled` + the family's resync |
| `d2b-provider-volume` | `Satisfied` off a `recover` marker, ignoring its own `converged` | the layout is re-derived through the idempotent ensure on adoption, and the binding set decides the outcome |
| `d2b-provider-volume-binding` | `Satisfied` while `socket_ready == false` (its own test pinned it) | `RetryScheduled` + the preserved resync |
| `d2b-provider-host`, `d2b-provider-user` | `Satisfied` over `Degraded`/`Pending`, no retry at all, and the cached-generation short-circuit pinned it | re-probe (`HOST_REOBSERVE`/`USER_REDISCOVER`), `RetryScheduled` |
| `d2b-provider-process` | `Satisfied` + `process-stopped` for a stopped lifecycle, before any probe or stop | the stop is established first; an unconfirmed stop defers |
| `d2b-provider-telemetry-{binding,service}` | `Satisfied` over `Degraded`/`Pending` once children existed | `RetryScheduled` + one re-check per pass |

**On the Cloud Hypervisor case specifically**, the concrete answer to your third point: the guest *does* query its dependencies (`dependencies_ready` over the spec's declared refs), and the bindings reach it through the controller session's fenced readiness — `BootstrapGraph.bindings`, gated by `bindings_ready`, which projects `Pending` while a binding's worker is not serving. Two gaps closed it: the pass now answers `RetryScheduled` with a 2 s re-check instead of claiming `Satisfied`, and — separately — `cloud_hypervisor_dependencies_ready` was reading devices, networks and volumes but **skipping `graph.bindings`**, so the Start/Restart precheck could pass with the binding the guest's volume service depends on still unready.

The generic half exists already and stays generic: `ctx.watch(target, WatchCondition::Ready)` with `ReconcileCause::DependentChanged`. The driver is still responsible for the case it cannot observe — which is exactly what the retry-after is for.

Every site's tests were corrected where they pinned the old claim instead of deleted, and the deferred shapes that had no test now have one. `make check` is 923/923 on this head; the full host lane is running.

PR #544, opened 2026-09-20
refactor(providers,d2bd): finish the landed families and declare operation rows in their crates

vicondoa:
# Legacy effect-port removal: enabling units and the two finishing units

Implements the first phase of `docs/plans/2026-09-19-001-refactor-legacy-effect-port-removal-plan.md`: the enabling units, the two families that already had a working envelope path, and the retirement of the declared-but-unbuilt surfaces.

## What this change does

- **Operation rows are declared in the declaring crate.** Each provider crate's `operations.json` is the source; the committed rows file and its five derived views are generated from the declarations plus the retained generic rows and pinned by the existing drift gate. A new parity gate checks the declaration against the crate's descriptor. The network and process families' 24 rows migrated; adding or renaming a family operation now edits only the declaring crate.
- **A provider-owned service carries the real payload.** The hosting fixture's byte-vector payload is gone: a service is invoked over the envelope's real request and response contract and receives a capability object built from its declared facets (driver context, declared state cells, per-zone kernel seam, declared descriptor legs). The composition root hosts a declared service through its registered factory and still refuses a declared service with no implementation.
- **The declared-but-unbuilt surfaces retire:** the two process providers' declared spawn ports and the transport crate's effect-port module, with their dossier, re-export, policy, and task-row pins. The live conformance launch port the supervisor backends use is untouched.
- **The process family is converted.** Its driver effects move out of `d2bd` (`process_effects.rs`, 1031 lines, deleted) into `packages/d2b-provider-process/` behind a declared service, built from host-supplied declared facets, with the composition-site injection removed.
- **The network family is converted.** Its kernel-invoking adapter and facets move into `packages/d2b-provider-network-local/` (`network_effect_port.rs`, 510 lines, deleted), reaching the same committed kernels with the same authority, and the resolved intent vocabulary moves to the crate that owns it.
- **The cross-binary proof grows.** The pilot gate now proves four operations end to end across the real broker and daemon binaries (`OpenPidfd`, `inspect-process-family`, `PollChildReaped`, `SeedDnsmasqLease`) with byte-for-byte operation names, fd carriage and liveness, and audit continuity keyed on the invocation id. Its fixture was also broken and is fixed (it pointed the broker at the public manifest and staged an ownership posture the verifier rejects), and its driver now selects the pidfd by kind because the forward carrier appends its own relay descriptors.
- **Review findings folded.** A hosted zone-plane method no longer takes its Zone from the caller's payload; startup refuses a method declaring a facet the host does not enforce; the network service builds its report through the canonical JSON path.

## Testing

- `make check` -> exit 0, 923/923 tests pass.
- `make test-host-integration` -> exit 0, 11/11 vmChecks pass (bridge-isolation, daemon-smoke, device-worker-launch, guest-agent-cap-confinement, guest-shell-service, privilege-oracle, resource-operator-activation, runtime-cloud-hypervisor-guest-preflight, state-posture-contract, virtiofsd-volume-runtime, wayland-proxy).
- Broker-seam pilot gate -> exit 0, four operations proven, audit continuity asserted.
- Focused crate tests and clippy run for every touched crate; the independent review's fixes each carry a red-then-green observation.

## Known residuals

Independent review of this diff (two reviewers, clean contexts) left these, none of which blocks the lane:

1. **The network family's bundle resolver is now a plane-composition snapshot** rather than re-loaded per reconcile. After the on-disk bundle is replaced, the family reconciles against the old intents and the broker's generation fence turns that into a permanent refusal until the daemon restarts; a tampered bundle no longer fails the reconcile closed per call. It matches the process and volume precedent, so it may be deliberate - it needs either a stated restart requirement or a per-invocation fresh fetch. **Owner decision.**
2. **Declared service-invocation facets are unenforced beyond the descriptor contracts.** Startup now refuses a declaration naming an unenforced facet, so nothing runs silently, but the enforcement path (caller identity, privileges, payload schema, deadline tier) is not built. Latent: both declared services use zone-plane methods carrying no facets.
3. **The hosted `has-active` and `inspect-network` services are unreachable through the only production dispatch**, which matches strictly on a declared operation; the declaration and dispatch shapes disagree until those methods become caller-facing.
4. **The zone-native bundle carries no NetworkManager unmanaged contract.** A fix that hardcoded it in the empty host model was reverted after review: the authority for a privileged host mutation belongs in the declared, hash-pinned bundle, and the empty model's fail-closed behaviour is the safety property. Consequence: the host-prepare path still fails closed for a bundle that declares no NM contract.
5. **The network fabric host-integration check is not in this change.** Its assertions were drafted (bridge, firewall projection, NetworkManager-unmanaged marking, routes, sysctls, hosts file, lease acknowledgement, restart adoption, finalize retirement) but it cannot reach them: the network controller's process launch is refused by the supervisor broker in that VM, so the bridges never appear. The same spawn path passes in the resource-operator-activation check, so it is the new check's VM posture. The nine host-mutating network operations are therefore unproven; the hermetic one is proven by the pilot gate.
6. **The two unrouted process surfaces remain:** three committed process wire rows (delegate-cgroup, open-cgroup-dir, launch-minijail-child) and the five-row process-systemd family still have no lane; both are recorded in the plan's Open Questions.

No additional operational monitoring required for this change: it is a refactor with no operator-visible behavior change by design, and it is validated by the two gates above (the hermetic cross-binary proof and the host-integration lane, which asserts live broker, daemon, and host posture).

Signals worth watching if a regression is suspected anyway:

- Daemon log terms: `provider registry refused`, `Zone authority index`, `stale-projection-generation`, `handler-refused`, `process launch failed` - the first two would indicate a bundle or registry problem, the last two a spawn-path problem.
- Expected healthy signals: the same launcher vocabulary, audit records, and refusals as before this change; one audit record per forwarded invocation on the pilot's operation set.
- Failure signal and rollback trigger: any operator-visible change in resource reconciliation, launcher vocabulary, audit shape, or refusal codes - the branch is revertible as a single squash if so.
- Validation window and owner: the first host integration run after merge; the lane author.


vicondoa:
**Retrospective independent review of head `308da861a55c9f366034e3f94a6735e7fa5212da`: partial, with one finding.**

The declared services, the dispatch, the ticket and facet vocabulary and the network kernel adapter all genuinely moved into the crates. What did not move is the substance: the launch, reconcile and finalize implementations remain in the daemon behind the runtime facets. So the effect-port surface moved while the behaviour it guards stayed daemon-side, and a summary describing the family as converted overstates it. Not a regression - the refusals and budgets survived - but the conversion is incomplete as described.

This review covers the diff as merged, because the change had landed without a review recorded against the head that merged. Bound to repository `vicondoa/d2b`, pull request #544, base branch `v3`, head `308da861a55c9f366034e3f94a6735e7fa5212da`.

PR #550, opened 2026-09-21
feat(providers): generate provider and service registrations from per-crate declarations

vicondoa:
# Generate provider and service registrations from per-crate declarations

A structural obstacle, and the change that removes it. The daemon's composition root registered each family's provider and service factory **by naming the family's crate**. That is fine for families the daemon already carries, but the layout policy forbids the daemon naming a family it does not carry, and the ratchet is shrink-only, so a new exemption row is exactly what a reintroduced daemon-side effect port looks like. A lane was blocked on it: the process-systemd family could not be registered without either weakening the ratchet or putting its declaration inside a sibling crate.

## What this change does

- **Each provider crate declares its own registration** in a `registrations.json` beside the declarations it already has: its provider and the services it serves. The two families registered by name today migrate onto it, the same way the row-declaration unit started with the landed families.
- **xtask renders the committed registration table** the daemon composes from, and pins it with a drift gate that fails a hand edit naming the artifact; generation is idempotent.
- **The parity gate refuses the shapes that would hide a mistake**: a declared service the crate does not spell as a const in its own sources, a service or provider declared by two crates (naming both), and a provider that is not the crate's own family.
- **The daemon's registration path names no family.** It iterates the table; a registered family the daemon does not yet carry starts with no drivers, and a declared service with no factory still refuses startup by name.

## The demonstration that matters

Adding a `registrations.json` to the blocked family's own crate made the daemon compose that family - visible in the startup order - with **no edit outside the declaring crate**, and the probe was then reverted byte-identically (the table's digest returned to its pre-probe value). That is the property the remaining lanes need.

## Testing

- `make check` -> exit 0 on this head (local Bazel profile).
- `make test-policy` -> 26/26, including the new parity/drift/idempotence gates.
- `cargo test -p xtask` all suites pass; `cargo check -p d2bd` clean; the pinned startup-order test updated and passing; the effect-service-factory-missing refusal by name still covered.
- The 30 `forward_rendezvous` failures in the `d2bd` cargo lib suite are pre-existing at the base commit (reproduced identically on a pristine `origin/v3` worktree).

## Known residuals

1. **Only the two landed families are on the table.** The remaining families migrate in their own lanes; until then a new family's registration is a declaration plus a factory, which is exactly what the blocked lane needs.
2. **The runtime driver wiring is still per-family in the daemon** (`registered_drivers`) - the table names no family, but the construction helpers are the families' own implementation, which is the layout ratchet's permanent dependency carve-out.
3. **`bazel build //packages/d2bd:d2bd_lib_test`** was reported failing to build in one agent's environment with crate-duplication errors in untouched files; `make check` builds and passes here.

No additional operational monitoring required: registrations are generated from declarations and the daemon's behaviour is unchanged for the families already carried.

- Signals to watch: a registered family starting with no drivers (expected only for a family the daemon does not yet carry), and the startup refusal naming a declared service with no factory.
- Expected healthy signals: the two landed families host their services exactly as before, and the startup order matches the pinned expectation.
- Rollback trigger: a service declared by a crate failing to host because its registration was not generated.
- Validation window and owner: the first family lane that adds its own registration; the lane author.


vicondoa:
**Retrospective independent review of head `4e7734848badf59b5d6e4dac89f8f8ff27284bd9`: partial, with one finding.**

The generated registration table drives iteration, but family dispatch and the service factories are still hand-written blocks beside it rather than being produced from the table. The design is held fail-closed by the catalog-mismatch check and the startup refusal for a declared service without a factory, so this is not a live defect. It does mean the table is not yet the single composition authority the change describes, which matters for whoever converts the remaining families.

This review covers the diff as merged, because the change had landed without a review recorded against the head that merged. Bound to repository `vicondoa/d2b`, pull request #550, base branch `v3`, head `4e7734848badf59b5d6e4dac89f8f8ff27284bd9`.

PR #551, opened 2026-09-21
refactor(host,d2bd): serve the host family's probe from its own crate

vicondoa:
# Serve the host family's probe from its own crate

The first Phase C family lane. The host family's probe moves into the declaring crate behind a declared service, and the daemon stops implementing it.

## What this change does

- **The probe lives in the crate.** `packages/d2b-provider-host/` carries the probe and the declared service; the descriptor builds its effects internally from declared facets, so the plane's construction site takes facets and no injected port appears there.
- **The daemon side retires.** The host half of `packages/d2bd/src/system_core_effects.rs` is deleted along with its injected descriptor arguments. The user family stays in that module until its own lane.
- **Observations are unchanged**, including the degraded observation published when a probe cannot complete; the parity evidence covers both the happy path and the degraded fallback with the same shape as the retired daemon implementation.
- **Restart adoption** is asserted: a restarted plane re-hosts the service from the same inputs and answers the same bounded observations.
- **The probe names no sibling provider's vocabulary.** Its reads are host state - kernel and OS identity, the caller's runtime sockets, device and module presence - and the two constants it previously imported from sibling provider crates are now local path and module names. The toolkit edge is the one a service-declaring crate legitimately needs, granted the same way the process and network crates are.

## Testing

- `make check` -> exit 0 (both before and after the dependency correction).
- `make test-policy` -> 26/26, with nine stale layout rows retired and two provider-side rows added (net inventory -7); no new shared/daemon family row.
- `cargo test -p d2b-provider-host` 19+4 pass, `-p d2b-provider-system-core` pass, `d2bd --lib resource_plane_v3` 37 pass including the hosted-service binding and the restart-adoption test; clippy clean on the touched crates.
- The host family serves no broker operations, so there is no cross-binary operation for the pilot gate to prove; the hermetic proof is the plane binding plus the crate's real-probe smoke test.

## Known residuals

1. **The user family is not in this lane** - it is a separate lane, and the shared module's row for it retires there.
2. **No KVM host scenario asserts the real host reads**; they are covered by the bounded real-probe smoke test and the plane binding test. A host-integration scenario for this family would extend the lane's own check, which is the pattern the fabric check will follow.
3. **The pre-existing guest-preflight registration wedge** is fixed in its own PR (#548); this lane's head rebases onto it after it merges.

No additional operational monitoring required: the probe's reads and the observation it publishes are unchanged, and the plane re-hosts the service on restart.

- Signals to watch: `HOST_EFFECTS_SERVICE` failing to host, or a host row publishing a different observation than before - the latter would mean the parity broke.
- Expected healthy signals: the same platform and capability observation, including the degraded one for a host that cannot be probed.
- Rollback trigger: a host row failing to reconcile because its service was not hosted.
- Validation window and owner: the next host-integration run; the lane author.


vicondoa:
Review evidence, rounds recorded against their heads.

**Round 1 (head `c9a72ac1b`):** correctness found no findings (it reproduced the retired daemon probe's semantics read-for-read); testing found a restart-adoption test comparing a live process count, which fails whenever any process starts or exits between its two calls; maintainability found the effects service diverging from the converted families, orphaned sibling vocabulary constants, a public constructor contradicting the crate's own claim, a stale test-support comment, and missing trailing newlines.

**Round 2 (head `034f5e433`, after the fixes):** correctness correct, no patch-introduced defect - the reshape preserves probe call order, the degraded fallback byte-for-byte, and error propagation; the constants pin compares two independently spelled constants rather than a re-export; the gate conversion is behaviour-identical for every input. Testing correct: no assertion weakened, and the scripted registry test still runs the real reconciler to a `Ready` end state rather than testing its double.

**Round 3 (head `cff90affa`):** one finding from round 2 remained actionable - a real-probe regression degenerating the process count to a constant would pass every test. The guard now asserts the field is present, well formed and at least one, and was proven by inducing the degeneracy: `panicked at resource_plane_v3.rs:3821:9: process count is degenerate: 0`. Delta review of that head is running; its verdict will be recorded before merge.

vicondoa:
**Delta review, head `ea04a1257`:** correct, one P3 that is cosmetic only - a missing space after a colon in a new comment (`check:the` at `probe.rs:278`). Recorded, not fixed: a whitespace-only edit would cost a head change, a fresh review and a full CI cycle for no behavioural or readability gain beyond one character, so it does not gate this merge.

Substantively the review verified what mattered: the guard **cannot flake** in its new home (the probe returns an error rather than zero when `/proc` is unreadable, so a successful enumeration on a running machine always counts at least the test's own process), and the guard is **complementary rather than redundant** across its three sites - the shape check panics only on missing or wrongly typed fields, so without the `>= 1` assertion a constant-zero regression would still pass the restart-adoption test silently. It also noted the one that must survive any future 'redundant assertion' pruning: the owning crate's guard is the only one that fails inside the producing crate's own suite.

vicondoa:
**Delta review, current head: correct, no findings, mergeable from this lens.**

**The exemption row is earned, and the citation settles it.** The trait is declared outside the crate - `packages/d2b-provider-system-core/src/host.rs:226` - and all three of its methods return `Result<_, SystemCoreError>`, so **any** implementation must name that type. The reviewer also established there is no narrower legitimate alternative: an alias re-export would evade the identifier probe only by hiding behind the import-line skip, which is precisely the evasion the ratchet's probes exist to defeat, and it would diverge from the production probe that names the type directly under the identical row pattern. The row keys on (crate, module, token), the schema's narrowest unit, shared by every existing row - and the stale-row direction of the check fails the gate if the signal ever disappears, so the row cannot linger after the reference goes.

**The rebase is honest.** The reviewer computed the incoming range (four commits, 18 files), the lane diff (70 files), and the **overlap between them: zero** - then compared per-commit tree pairs and found lane content byte-identical across the replay, so no silent conflict resolution could have dropped content on either side.

**The spacing fix is one character** with no reflow of the surrounding comment.

Residuals recorded: the layout gate itself was not executed because it is serialized under the repository lock, so the reviewer replicated the probe logic directly against the tree and reproduced exactly the seven reported signals; and the generator's idempotency was argued from disjoint file sets rather than re-run, which holds because the incoming and lane files do not intersect.

PR #554, opened 2026-09-21
test(vsock): drive the transport deadline tests on virtual time

vicondoa:
## What

The two open-path deadline tests raced real sleeps against a real-clock deadline. A starved process could satisfy the fake's sleep before the enforcing timeout fired, because `tokio::time::Timeout::poll` polls the inner future before checking its own delay - so under load `open` returned `Ok` and `unwrap_err()` panicked. The endpoint-close grace-period test raced the same way and is converted with them.

The deadline now sits on tokio's clock, so it shares the clock the enforcing timers use (a `std` deadline desynchronises from them under a paused clock). Production behaviour is unchanged on an unpaused runtime; the effect trait's deadline parameter moves with it, and its only in-tree implementor is the test fake. The tests run on a paused runtime and are driven by stepping the virtual clock, so they no longer read wall time at all.

Assertions are unchanged: both open tests still assert `DeadlineExceeded`, and the close test still asserts a completed close with `TransportPhase::Released`.

## Validation evidence

- **Exact `make check`:** result=`passed` (923 of 923 tests), run plain with `D2B_BAZEL_JOBS=4`, no profile override.
- **Pre-fix reproduction:** 3 failures, including the third same-class test found during reproduction. Gathered under manufactured CPU load, and labelled as such.
- **Post-fix:** passes in 0.6s, twice.
- **Ten consecutive runs under ambient load:** passed - max 0.6s, min 0.5s, dev 0.0s.
- **Induced failure:** `remaining_until` was temporarily patched to return 60s so the deadline could never fire; both deadline tests failed with the exact original symptom (8 passed / 2 failed). The patch was then reverted and the target re-verified green.
- **Changelog:** fragment added under `changelog.d/`, since the manifest change is code-visible.

## Review


vicondoa:
Review evidence, bound to head `ea25cf365` and observed base `v3` at `310eeaaf0`.

**correctness-reviewer:** verdict correct, no findings. Independently verified there is exactly one in-tree implementor of the effect trait and one call site (not accepted from the description); that both deadline tests and the close test keep their assertions verbatim against the pre-change versions; that the virtual-clock driver is bounded and cannot skip the assertion; and that the trait-signature change is a semver break for an out-of-tree implementor, recorded as a residual.

**testing-reviewer:** verdict correct, no findings. Proved no silent-pass path exists, from the clock ordering itself - the fake's sleep registers at 1100ms of virtual time against a 1000ms deadline timer on the same clock, so the sleep cannot win; confirmed no residue of the induced-failure experiment; confirmed the `test-util` dev-dependency is scoped to dev-dependencies.

Merged as `e06915d47` by squash with an expected-head guard on that head.

PR #555, opened 2026-09-21
test(xtask): wait for the facade barrier instead of five seconds

vicondoa:
## What

The facade-isolation test in the BuildBuddy configuration suite held a fixed five-second wall-clock deadline and hoped every facade arrived inside it, so a busy enough machine could fail the test for being slow.

It now waits for each facade to reach its barrier, with an exited child as the only liveness bound - a child that has exited can never produce the marker, so the loop still terminates while holding no wall-clock guard. The poll interval remains; the deadline does not.

## What this does and does not claim

The five-second window was **not** observed to fire: this target passed in sixty-plus runs, including two full suite runs under load in which six other targets flaked, and in the final `make check` (0.4s). So this removes a dependence on wall time rather than repairing an observed failure, and the evidence is structural rather than a reproduction:

- **Mechanism:** before, `for _ in 0..500 { if exists { break } sleep(10ms) } assert!(exists)` - five seconds of hoping. After, `loop { if exists { break } if child exited { panic } sleep(10ms) }` - progress-bounded, no wall-clock guard anywhere in the test path.
- **Induced failure:** a facade forced to exit before reaching the barrier fails the test with `warning facade exited with exit status: 1 before reaching the barrier` in 0.47s; with the warning suppressed, the existing `assert_eq!` catches it (left `Some(0)` / right `Some(1)`).
- **Ten consecutive runs under ambient load:** 10/10 passed, against continuous fleet build activity.

## Validation evidence

- **Exact `make check`:** 920 of 923 passed, plain, `D2B_BAZEL_JOBS=4`, no profile override. The three failures are outside this change and observed in other lanes: the credential lease timing test, the broker reaper timing tests, and the meta tier0 timeout on a vanished spawn-runner directory. All three are being fixed in their own lanes.

## Review


vicondoa:
Review evidence, bound to head `2cae32ed9` and observed base `v3`.

**testing-reviewer (round 1):** correct, one P3 - the exited-child panic path returned without releasing the facade wrapper or reaping the child. Fixed: the panic path now publishes the release marker and reaps before failing, proven by inducing a facade exit before the barrier and capturing `warning facade exited with exit status: 1 before reaching the barrier`.

**correctness-reviewer (round 1):** correct, no findings. Traced both sides of the process boundary and established that the hang the removed five-second bound guarded against is **not reachable in-script**: the child runs the wrapper in the foreground with an inherited environment, and every pre-marker step either completes or exits - so the bound was guarding nothing and dropping it costs nothing real.

**testing-reviewer (round 2, head `2cae32ed9`):** correct, no findings. Verified the delta is exactly the claimed six insertions with nothing else moved; that the release write cannot mask a genuine failure because it is reached only after the failure branch and is unconditionally followed by the panic; and - reading the pinned toolchain's own `std` source - that `try_wait` caches the observed status, so the subsequent `wait()` cannot block.

PR #556, opened 2026-09-21
test(broker): freeze the audit write window in tests

vicondoa:
## What

The broker's audit write-limiter tests need several writes to land inside one rate-limit window, but that window rolled over on the real clock. A machine slow enough to spend longer than one window on the preceding writes would roll the window between writes, let an excess write through, and fail an assertion that is really about the limiter's policy rather than about timing.

The window is now frozen in tests through a `cfg(test)` seam on the limiter. Production is untouched: `window_frozen` does not exist outside test builds, `freeze_windows` is `cfg(test)`, and it is reachable only from the `cfg(test)` command whose sole caller is the test helper.

## Validation evidence

- **Mechanism:** before, `check` rolled the window with `if self.window_start.elapsed() >= AUDIT_WRITE_WINDOW { reseat }`; after, `window_elapsed()` returns false when the window is frozen and otherwise makes the identical real-clock comparison. Release builds take the unchanged path.
- **Induced failure:** with the limiter temporarily patched to never refuse an unprivileged write, both named tests fail exactly as designed (`second unprivileged refusal must be dropped`, and the suppressed-warning counter test), and the queue-full sibling still passes because its drop happens caller-side. The probe was reverted; the diff carries no residue.
- **Thirty consecutive runs under ambient load:** 30/30 across the five write-limit tests while roughly a dozen sibling lanes were building.
- **Honest limitation:** the rollover was **not** reproduced on demand - both named tests passed all ten pre-fix runs even under manufactured load before that instruction was retracted - so the proof is the mechanism plus the induced failure, not a reproduction.
- **Exact `make check`:** 921 of 923 passed, plain with `D2B_BAZEL_JOBS=4`, no profile override. The two failures are known sibling flakes owned by other lanes (a reaper registration race and a route-test assertion); all five audit write-limit tests pass inside both binaries.
- **Siblings left alone:** three other write-limit tests were checked and genuinely have no window dependence, so they are unchanged rather than mechanically edited.

## Review


vicondoa:
Review evidence, bound to head `67feab7e6`.

**correctness-reviewer:** correct, one finding. Production is genuinely unchanged - it verified every caller of the `cfg(test)` seam itself, down to confirming that the one non-`audit.rs` caller sits inside a `#[test] fn` - and the finding is that the window-rollover branch is now exercised by **zero** tests.

**testing-reviewer:** correct, one finding, the same coverage gap reached independently and at confidence 75, plus the reason it is permanent: freezing is applied inside the shared test constructor, so every test inherits frozen time and there is no unfreeze path. It also confirmed the five write-limit tests keep their assertions verbatim and that no experiment residue remains.

Fix in progress: freezing becomes a per-test opt-in and the rollover branch gets a deterministic test, proven by making the reseat a no-op and watching the new test fail. Verdict for the new head will be recorded before merge.

PR #557, opened 2026-09-21
refactor(user,d2bd): serve the user family's effects from its own crate

vicondoa:
## What

The user family's effects move out of the daemon and into `d2b-provider-user`: the crate now declares its own effects service (`user.d2bus.org/effects`), its facets, its probe and its registration row, and the daemon composes it through the generated registration table rather than naming the family.

The daemon-side module that hosted it is reduced to the **host** half, which is deliberately left intact so this lane and the host-family lane do not collide; whoever merges second rebases. One sanctioned visibility widening: `packages/d2b-provider-toolkit/BUILD.bazel` gains the user crate as an enumerated consumer, matching the edge the process, network-local and host families already have, because a service-declaring crate needs it.

## Scope discipline

66 files, +1234/-220, all user-family. The diff against `origin/v3` is **zero** in every path other than this lane's, with the work of #547-#552 and #554 preserved byte-for-byte - so this cannot revert another lane's change. A first refresh attempt produced exactly that hazard, which is why the diff was checked path by path.

## Validation evidence

- **Exact `make check`:** exit 2 with exactly three failures, all in the known load-flaky broker suite being fixed in other lanes (`d2b_broker_test`, `d2b_broker_fake_backends_test`, `d2b_broker_layer1_bootstrap_test`). Re-run in isolation with the identical command and the same cap: **3/3 PASSED (exit 0)**. No test was changed, no retry added, no assertion weakened.
- **Focused gates:** user crate lib suites, system-core, and the daemon plane tests (inspect-user binding, restart rehost, committed startup order, started-provider set) all pass on the rebuilt tree.
- **Generated artifacts through the authority:** `make generate` exit 0 and idempotent - policy-input closures, metadata and lockfiles regenerated, and the registration table regenerated via `cargo xtask check-provider-crate-layout --fix`, with the user row present. Both re-runs produced no further diff.
- **Changelog:** fragment updated for the user half only, matching the scope above.

## Review


vicondoa:
Review evidence, bound to head `003c3790`.

**correctness-reviewer:** correct, four findings - the service is generic with a test-only constructor where every converted sibling is concrete over a facet-carried trait object, and the declared facet boundary is *discarded* (the constructor takes the facets and drops them), so the composition seam and the scripting seam are two different surfaces; a changelog bullet glued onto the previous line; six files missing trailing EOF newlines; and documentation the move falsified - a crate README still claiming the daemon implements the effects and that the crate cannot reach host state except through its port, while the crate now opens NSS itself.

**testing-reviewer:** correct, two findings - the same shape divergence, with the added detail that the scripting double lives in a crate-private `cfg(test)` module rather than the release-gated `test_support`, so the plane and integration tests cannot script the probe hermetically; and a new registry test that can **silently skip all of its assertions** depending on the machine's account name.

Fix pass in progress for all six items. Verdict for the new head will be recorded before merge.

vicondoa:
**Verification of the current head: correct.** The corrected field doc is accurate - it matches the production facet construction, the crate's own bounded probe, the inline construction-site comment, and the sibling facet field docs. The delta is documentation-only: seven changed lines, all comments, one file, no code movement.

**On the exclusions you claimed**, which is the substantive part of this review: two hold and one does not.
- The changelog exclusion holds: that fragment describes the user move as a past transition that also matches the present state.
- The interaction-family exclusion holds: the text cited is explicitly scoped to child sources whose effects still live daemon-side.
- **One does not, and it is deliberately left alone.** The section banner in `shared_provider_effects.rs` claiming "the daemon implements each declared family's port" is an unscoped universal that the user and process conversions falsify - the user family's discovery port is implemented crate-side now, with no daemon implementation. But the text **predates this branch's base** and is already contradicted in-tree by a corrective comment two lines beneath it, so it is a pre-existing stale sentence rather than anything this PR introduced. It is recorded as such, with the one-line rewording a future pass could apply ("the remaining families' ports").

Confidence on that item is 55 and it is pre-existing, so it does not gate this merge; the finding is recorded rather than cycled, which is the same treatment every pre-existing observation has had today.

vicondoa:
`631e65c5283d2883aae0c78fecd6d3af300f11aa`.

**Independent review: correct - one finding being closed before merge.**

The conversion is real rather than a shim, which is the thing worth stating plainly given how often that is not true: the daemon's user discovery and its production effect implementation are deleted, no reference to the user seams survives anywhere in the daemon, and the probe moved byte-for-byte - same digest domain, same bounded name-service reads and group checks - into the provider crate, which is what production composition now wires. Composition runs through the generated registration table exactly once, with both table consumers handling the row and the hand-written composition site removed. Refusals survive: an absent account still reports the same pending classification, a discovery failure keeps its retryable classification, and the new hosted surface adds its own closed refusal codes. The boundary change is a single visibility grant matching the new dependency edge, with no cycle. The moved tests were strengthened rather than weakened - a call-order assertion became a behavioural assertion over the facet-carried probe, and restart re-hosting is now covered.

One finding, being closed:

**P3 - the hosted inspect-user surface answers over different inputs than it documents.** It synthesises its input from the request username alone, and that minimal spec declares no groups. The probe's identity digest is derived partly from the declared group names, and group memberships are only required when the spec declares them. So for a row whose stored specification does declare groups, the hosted surface returns an identity digest that differs from the digest the driver publishes for that same row, and can report the account discovered and ready where the row itself classifies it drifted - a declared membership that does not verify is invisible to this path. The method is documented as answering the same probe and classification the driver reconciles over for that declared identity, and that holds for the mechanism but not for the inputs. It is latent: no production caller consumes the method and the existing tests use group-free identities, which is why they cannot detect it. Being closed so that the tree does not assert an equivalence it does not have.

---

Review bound to repository `vicondoa/d2b`, pull request #557, base branch `v3`, head

vicondoa:
**Re-verification of head `6c972a46b9803403836c77cb11eec4b8fcfb2369`: clean - the finding is resolved.**

The request field carrying the declared identity is genuinely required rather than defaulted: parsing refuses a missing or mistyped field, validates every element through the group-name parser so empty, oversized and malformed entries refuse, and bounds the count - there is no defaulting path. The hosted surface builds the served specification from the declared identity, and since the probe reads only the operating-system username and the declared groups, the probe input, the identity digest and the required bindings are the driver s own for the same row - so the two answers cannot diverge. And the new test fails on regression rather than merely passing now: it pins the row s group-inclusive digest as different from the group-free digest and requires the hosted answer to equal the row s own drifted observation with that same digest, so a return to the minimal specification would flip the answer to ready with the group-free digest and fail all three assertions.

---

Review bound to repository `vicondoa/d2b`, pull request #557, base branch `v3`, head `6c972a46b9803403836c77cb11eec4b8fcfb2369`.

vicondoa:
**Re-verification of head `07c48b849c1a74b6709824d671ebf394126ffb7c`: clean - everything clear.**

The rebase resolution was verified by presence. Both families arms coexist rather than one replacing the other, in the service-factory map, the driver map, the test inputs and both test groups; the startup-order test lists the incoming family in the table-order prefix and out of the unconverted tail, which is what makes the table authoritative rather than aspirational.

The shared daemon module that each conversion half-emptied is deleted consistently rather than merely removed: it is gone with no references anywhere in the daemon or the policy tooling, all ten ratchet rows naming it are retired along with the module registration, and both families now start through the registration loop with no hand-written declaration remaining anywhere - which is the condition under which the table is genuinely the composition authority.

The fix this review required survived the rebase byte-for-byte, confirmed by comparing file blobs across the rebase rather than by trusting it: the required validated groups field, the specification built from the declared identity, and the regression test are all intact.

---

Review bound to repository `vicondoa/d2b`, pull request #557, base branch `v3`, head `07c48b849c1a74b6709824d671ebf394126ffb7c`.

vicondoa:

PR #558, opened 2026-09-21
test(cli,bus,providers): remove wall-clock dependence from timing tests

vicondoa:
## What

Seven timing-dependent test files stop depending on wall-clock behaviour, converted by failure class rather than mechanically:

- **Upper bounds replaced by causality.** `packages/d2b/tests/audit_contract.rs` asserted the CLI finished within ten seconds of its own one-second deadline. It now asserts the deadline actually fired - exit code 1 plus the named `deadline-exceeded` refusal - with a generous watchdog that only converts a hang into a clean kill. The old bound could only false-fail under load, and could never catch a hang, because `output()` blocks before the assertion is even reached.
- **`packages/d2b/src/context.rs`** dropped an `elapsed < 5s` check, keeping the `TimedOut` error kind as the mechanism that proves the receive budget fired.
- **Deadline races converted to virtual time.** In `packages/d2b-bus`, the operation's 500-tick deadline counted real time since bus creation under the system clock, so under load the deadline expired during setup before dispatch ever happened - that is the actual cause of the suite's flake, not a slow machine. It now runs against an injected `ManualClock` with virtual-time guards, so the deadline is deterministic.
- **Fixed-sleep-then-assert converted to condition-driven waits** in the credential binding, guest controller and supervisor adapter test paths.

## Validation evidence

- **Every owning target passes 10 consecutive runs under ambient load** (no synthetic load was generated).
- **Exact `make check`: 923 of 923 tests pass, exit 0**, run plain with `D2B_BAZEL_JOBS=4` and no profile override.
- **Induced-failure experiments** show each converted negative assertion still catches its bug, rather than passing because it no longer checks anything.

## Integration note

`packages/d2b-provider-supervisor/src/adapter.rs` is also being edited by the retrospective fix lane, which changes production semantics there (a dropped deadline registration and a readiness-probe classification). This change touches only that file's test `wait_until` helper. Whoever merges second rebases; the overlap is small.

## Review


vicondoa:
Review evidence, bound to head `a8f454aee`.

**testing-reviewer:** correct, two findings, and it corrected the doctrine this PR was written under. Removing an elapsed bound is not uniformly right: the bound it removed in the context module sat over a 100ms receive budget - fifty times headroom - so it was never plausibly the flake source, and its removal means a regression that reads that budget as seconds now passes with the right error kind after taking a hundred times as long. The second finding is the same shape in the audit guard, where a deadline inflated into the 10-30s band slips inside the watchdog.

**correctness-reviewer:** correct, no findings at any severity. It verified the manual-clock injection is sound (production still gets the system clock through the same constructor; the deadline fires deterministically with ticks derived from the operation rather than fitted to the assertion), that no watchdog is reachable on a loaded machine or dead code, that the audit test is not vacuous, and that the supervisor adapter change is confined to the test module. It added a pointed residual: the comment claiming "the harness timeout fails" a hang is false for the in-repo lane, because the nextest configuration sets no per-test timeout.

Fix pass in progress: generous-but-finite bounds restored where coverage was dropped, with headroom stated per bound. Verdict for the new head will be recorded before merge.

vicondoa:

PR #559, opened 2026-09-21
fix(d2bd): log the typed bundle-load error instead of discarding it

vicondoa:
## What

The per-invocation bundle loader was written as `crate::load_bundle_resolver(&daemon_state).ok()`, which discards a typed error that distinguishes a `BundleTampered { path, reason }` from a `InternalIo { context, detail }`. Any bundle problem therefore produced the same generic broker refusal with nothing in the journal to say why. The closure now logs the typed error where it collapses, and still returns no intent - the fail-closed posture is byte-identical, only the reason is now recoverable.

This was found by a retrospective review of the six pull requests that merged before this repository required review on every PR, and it is a **regression**: the adapter that the per-invocation loader replaced logged the typed per-call refusal (`network_effect_port.rs` at `612fd21dd`), which the conversion dropped.

Chosen deliberately the smaller option: carrying the error kind into the refusal detail would have changed a cross-crate trait contract (the provider crate's intent-source loader signature and every resolve method) for a diagnosability fix, so the log line is the honest minimum.

## Validation evidence

- **Test:** `shared_provider_effects::tests::intent_loader_logs_the_typed_error_when_the_bundle_is_tampered` tampers the on-disk bundle (drops `bundleHash`, yielding `BundleError::Tampered { reason: "missing-bundle-hash" }`) and asserts both that no intent is produced and that the typed error reaches the journal.
- **Pre-fix proof:** with the closure reverted to `.ok()`, the test fails at the journal assertion with an empty capture while the fail-closed assertion still passes - so the test catches exactly the regression it describes.
- **Sibling behaviour unchanged:** `network_effects_observe_a_replaced_bundle_without_a_daemon_restart` still passes, confirming per-invocation reload is untouched.
- **Exact `make check`:** a first attempt aborted before running any test because the remote build service was unresolvable (`UNAVAILABLE: Unable to resolve host d2b.buildbuddy.io`); it was not retried or worked around. Re-run on this head is in progress and the result will be recorded here.

## Review


vicondoa:
Review evidence, bound to head `e944c9128` and observed base `v3`.

**correctness-reviewer:** verdict correct, no findings at any severity. On the redaction question - this change logs an error that carries a bundle path - it established that printing that path is **already sanctioned by the repository's own precedent**: the same field is logged for the same variant in the typed-error path, the contracts crate deliberately exempts this variant from its no-path assertion because its message carries the artifact path for operator diagnostics, and the broker already writes that path into its audit log. No secret, hostname, or identifier is involved, and the reason is a fixed slug. It also verified the fail-closed posture is byte-identical - a log call inserted before the existing `None` return, nothing else.

**testing-reviewer:** verdict correct, no findings. Verified the journal capture is genuine (a real tracing subscriber writing to a buffer, with the loader closure running synchronously on the caller's thread so the multi-threaded runtime cannot race it), that the on-disk tamper produces exactly the asserted typed error, and that the assertion pins a stable variant identity rather than incidental prose.

Two notes below reporting threshold, recorded rather than acted on: two `:the` typos in the new test's doc comments, and the new changelog file lacking a trailing newline.

vicondoa:
**CI gate failure and fix, recorded for the head that follows.**

`//bazel/checks/policy:check-async-gate` failed on the reviewed head in CI and reproduced locally (919 of 923 passed, this gate failing in 8.1s). The gate's own output named it exactly:

```json
{"api":"std::fs::write","code":"std::fs::write(","context":"async fn","file":"packages/d2bd/src/shared_provider_effects.rs","line":3390,"violation":"blocking-call-in-async-context"}
```

The new test is async and tampered the bundle with the blocking write. The async-purity gate flags deny-listed calls in async contexts and an allow attribute does not exempt them, so the fix is the sanctioned async equivalent (`tokio::fs::write(...).await`), which changes the call mechanism and nothing else.

Verified after the fix: `check-async-gate` **PASSED** (exit 0), and the test itself **PASSED** (1 passed, 0 failed, 485 filtered) - it still observes the typed tamper error through the same journal capture. A delta review of the new head is running; its verdict will be recorded here before merge.

vicondoa:
**Independent review of head `3650920baf55719d603a1d01410607151e699c4b`: correct - no findings.**

The change replaces a discarded error with a logged one and nothing else, which is the property that matters for a diagnosability patch: the set of reachable outcomes is identical to the previous form, so the fail-closed posture and the caller s generic refusal are unchanged and only the reason is added. The diagnostic is actionable rather than merely different - a tampered bundle and a missing or unreadable one produce distinguishable variants carrying the path and the reason, so whoever reads the journal can tell which happened.

The test covers the claim rather than the occurrence: it drives the patched construction path, tampers the stored bundle so it fails verification, asserts both that no intent is produced and that the journal actually carries the typed variant through a real logging capture. Reverting the change empties that capture and fails the assertion while the fail-closed assertion still passes, which is exactly the shape a diagnosability test needs - it can only pass if the reason survives to the log.

---

Review bound to repository `vicondoa/d2b`, pull request #559, base branch `v3`, head `3650920baf55719d603a1d01410607151e699c4b`.

PR #560, opened 2026-09-21
test(resource-runtime): assert the mechanism instead of sleeping first

vicondoa:
## What

Eight timing sites in `d2b-resource-runtime` stop sleeping before asserting. They were the worst class of timing dependence in the tree, because a sleep-then-assert-negative test does not merely flake under load - it **silently passes when the behaviour under test is broken**. A slow mailbox made "no extra reconcile happened" vacuous; a requeue timer that was never cancelled was invisible because the actor that would have processed it had already stopped.

Converted with two mechanisms, chosen per site: **causality** where a subsequent observable event proves the earlier work completed (a barrier status, a watch notification, a `wait_status` on the terminal outcome), and **virtual time** (`start_paused` plus a `pass_virtual` helper) where the test is about a timer deadline.

The cancellation test needed more than a mechanism swap. Its subject is a stale requeue timer that must *not* fire, and the old version could never observe the bug: the deleting actor is stopped, so an uncancelled timer's message is dropped unprocessed. It now keeps the deleting actor alive across the requeue deadline via a test knob, advances the clock past it, and asserts the absence of a second delete on the delivered signal rather than on elapsed time.

## Validation evidence

- Each converted site is asserted on a signal and cannot pass by advancing nothing: where the clock is virtual, the guard branch panics rather than satisfying the assertion.
- Focused target runs and per-site before/after analysis are in the lane's report; the aggregate gate runs under the repository-wide gate lock, since a starved scheduler produces NO STATUS rather than results.
- Changelog fragment added.

## Review


vicondoa:
**Verdict: correct, no patch-introduced defects.**

**The direct-message cancellation is honest**, which was the question worth asking: the requeue cancellation lives entirely in the actor's delete handler, and the manager's remove RPC contributes only the durable deleting-mark commit plus a re-send of the *same* message. So the direct send exercises identical cancellation code, and what it skips - admission, child recursion, mark commit, status publication - does not interact with timer cancellation, because the invariant under test routes purely on the actor-local deleting flag. The full production entry point still runs end to end in the completion phase, which is where the mark is committed in production order.

**The assertion reduction strengthened coverage rather than weakening it.** The removed tolerant check is now covered strictly, and the stale-timer signal has *moved* to the exact delete-count assertion, which is stronger than the reconcile-count assertion it replaced: a surviving requeue re-drives the delete pass and the count becomes two, whereas the old assertion could only catch a stale reconcile processed after the actor had already stopped - that is, almost nothing. The surviving reconcile assertion is labelled with what it does not prove.

**The author's flake reasoning holds against the code**: the remove RPC awaits an off-thread store write, and that idle window is where the auto-advance fires and re-arms the requeue, matching the measured spread that tracked store-I/O duration. Deferring the client call to the completion phase is sound because by then a terminal verdict schedules nothing.

Two residuals below the finding bar, recorded: the direct-send window runs a delete without the durable mark, which production never does and which the actor-local invariant does not depend on; and the setup margin relies on the status publication preceding requeue arming, so a future reversal of that ordering would surface deterministically rather than silently.

PR #562, opened 2026-09-21
test(xtask): resolve workspace metadata without the registry

vicondoa:
## What

The provider-crate layout matrix test shelled out to `cargo metadata` with a fresh empty `CARGO_HOME`, a runfiles tree that shipped **no `Cargo.lock`**, and neither `--locked` nor `--offline` - and discarded cargo's stderr, so any failure surfaced as the fixed string `provider-crate-layout-metadata-failed`.

Under load that produced two symptoms from one cause: a fast non-zero exit (panic with that static string), or the whole 404-test suite crawling past its budget with every test logging "has been running for over 60 seconds". The identical binary passed the full suite in 50.53s once the machine quieted - which is exactly what made an unhermetic subprocess look like a load flake.

Now: `--locked` always, `--offline` first with an online retry, the lockfile added to the test's runfiles, and cargo's own stderr surfaced in the failure string. Assertions, timing, and the test's scope are untouched.

## Why this shape

Not invented here: the production-closure test performs the same operation the same way - locked, offline-first, with stderr - and does not flake. This adopts the repository's existing pattern rather than introducing a new one.

## Validation evidence

- The new invocation with a fresh empty `CARGO_HOME` exits in **1.02s**; the previous form took 11.1s offline-without-lock and 45.3s against a dead proxy, so the registry and network dependence is gone.
- The **identical full 404-test suite passes in 50.53s under load 44-63**, including the previously failing test, where the same binary had timed out three times at 300.3s under load 100-120. That separates the fix's effect from machine starvation.
- Pre-fix failure text captured verbatim from three consecutive failing runs.
- A competing hypothesis was tested and **disproved**: this crate has no recursive walker - its tree checks are shallow `read_dir` calls - so an earlier "unbounded walk" theory does not apply to this test.

## Review


vicondoa:
**Correction to this PR's rationale, from review - the fix works, my explanation of it was wrong.**

The body above credits `--locked` and the lockfile for the hermeticity. A reviewer verified empirically that this is not what does it: `cargo metadata --no-deps --locked --offline` **succeeds with a missing or stale `Cargo.lock`**, because `--no-deps` skips dependency-graph resolution entirely, so `--locked` has nothing to enforce. The real source of the hermeticity is `--no-deps`: no graph resolution means no registry access at all, which is what the timing evidence shows (1.02s against a fresh empty `CARGO_HOME`, versus 11.1s offline-without-lock and 45.3s against a dead proxy before).

Two consequences, both being corrected in code: the comment claiming the lockfile pins the graph is inaccurate for this invocation, and the `//:Cargo.lock` entry added to the test's runfiles is never read by the check as invoked - dead input data that reads as load-bearing.

What the correction does **not** change: the fix is sound and the test is not blinded by any of this. The matrix comparison reads the metadata listing, which is unchanged, and lockfile drift is covered by the separate production-closure checks rather than here. The duration evidence stands.

PR #563, opened 2026-09-21
docs(providers): state the converted family shape as a requirement

vicondoa:
## What

Every family conversion in this program has had to re-derive its shape by reading the merged siblings, and two consecutive review rounds found the **same divergences twice** - a generic service with a second test-only constructor, and a facet boundary that was declared and then discarded. An unpinned convention recurs, so this writes it down where the layout reference already lives.

The new section covers:

- **What a converted crate carries**: the effect port trait, the concrete service holding a facet-carried trait object, the registration surface, and the scripting double behind the release gate so other crates can opt in.
- **One seam**: production and tests cross the same trait object through the same constructor and descriptor, and the daemon realizes the port at the family's registered row rather than naming a family branch.
- **What a reviewer rejects**, with the reason attached to each - the five antipatterns that actually occurred: a generic service with a second test-only constructor, a discarded facet boundary, a double that is not release-gated, a registration or binding test that reads the real machine, and prose left describing the old shape.
- **A conversion completes when**: generated views came through the authority, a changelog fragment exists, every README or doc comment the move invalidated is updated, and added files end in a trailing newline.

Naming the antipatterns is the point: a rule without its failure mode is what gets ignored, and each of these five cost a review round to find.

## Validation evidence

- The document is hand-written committed prose, not generated - verified by checking that no generator emits it, so no regeneration applies.
- Pure ASCII and ends with a trailing newline.
- A separate hypothesis was tested and disproved during this work: the gate failures seen in the same window were timing walls and timeout caps in the broker, xtask and layout components under fleet starvation, which a single markdown file cannot affect.

## Review


vicondoa:
**Verdict: correct, no findings.** Every corrected claim was verified against the merged process and network-local conversions and against the composition root - the crate-side effect implementation over daemon-supplied facets, the facet-versus-implementation distinction, the weaker and true composition rule, and the corrected completion checks. The document now describes the shape the tree actually has, which is the only thing that makes it usable as a rejection standard.

vicondoa:
`a0507736b51cae348954ee5c300149e59d37f13e`.

**Independent review: incorrect - two findings.**

Both are of the kind this document exists to prevent, so they matter more here than they would in code: a layout reference that diverges from the tree propagates the divergence to every lane that follows it.

**P2 - the corrected sentence claims a state the tree does not have.** It states unconditionally, in the present tense, that each family crate serves its own effect implementation over the declared facets. That holds for the families that declare effect services and are hosted through the generated table - host, network-local and process. For the remaining family crates the production implementation of the declared port still lives in the daemon: guest (`d2bd/src/guest_effects.rs:1461`), credential (`d2bd/src/credential_effects.rs:65`, a substantive implementation over the daemon's session registry), activation, binding, endpoint, user (`d2bd/src/system_core_effects.rs:28`), device and usbip (the shared provider effects), interaction and volume. Their shipped prose says the same thing - the security-key crate's README describes the port as one the daemon implements, and the volume crate's says the crate carries no effect implementation.

So a later lane reading that sentence will believe every family crate already serves its own implementation, and a reviewer applying this document's rejection list to those families will reject shapes the tree legitimately still carries. Scope the sentence to the families that declare effect services - as the section's own heading already does.

**P3 - the registration bullet names a type that does not exist.** The declaration artifact is given as `<Family>Descriptor`. The declaration the plane registers is the shared `d2b_resource_types::DriverDescriptor`, built by a per-family constructor (`credential_descriptor`, `host_descriptor`, `volume_descriptor`), which the crates' own READMEs record correctly. Since the neighbouring parentheticals in the same bullet are all real type names, a reader will go looking for a `CredentialDescriptor`-style type and not find it. State the declaration as the shared `DriverDescriptor` built by the `<family>_descriptor()` constructor.

---

Review bound to repository `vicondoa/d2b`, pull request #563, base branch `v3`, head

vicondoa:
**Re-verification of head `27a2059e138a815fc226dd9e26b17fcbefc306f7`: correct - both findings cleared, count now right.**

The count was verified rather than accepted, four independent ways per family: exactly four crates carry a registration declaring a service and no fifth exists, each of the four implements the family effects trait inside its own crate, the generated registration table carries exactly those four rows, and no implementation of any of the four ports remains anywhere in the daemon. The contrast clause holds too - every crate the sentence says still implements its declared port in the daemon has a real implementation site at this revision, and none of the four named crates appears among them.

That is the property this document needed: a reader can now tell which families serve their own effects and which do not, and can check either claim against the tree without trusting the sentence.

---

Review bound to repository `vicondoa/d2b`, pull request #563, base branch `v3`, head `27a2059e138a815fc226dd9e26b17fcbefc306f7`.

PR #564, opened 2026-09-21
refactor(guest,d2bd): serve the guest runtime family from its own crate

vicondoa:
## What

The guest runtime family's effects move out of the daemon and into `d2b-provider-guest`. The 1758-line `packages/d2bd/src/guest_effects.rs` is deleted; its content moves into the crate as a concrete effects service over declared facets, with its own registration and a scripting double in the release-gated `test_support`.

Two facets carry what the family genuinely needs from the daemon: a manager view (row view, committed provider identity, controller-session generation) and the Cloud Hypervisor runtime. The daemon keeps only those implementations, built in its composition root with the existing fails-closed locking semantics preserved.

The daemon now starts the family through the generated registration table - one `registered_drivers` arm and one service-factory arm - with the hand-written family site deleted, matching the process, network-local and host precedent. The controller state machines for the qemu-media, azure-container-apps and azure-virtual-machine providers move unchanged.

## Validation evidence

- Focused verification green: the guest crate's tests, the plane module tests (including a new binding test for the family's effects service), and the layout authority.
- The aggregate gate was not run locally: it is serialized under the repository-wide lock and CI gates the rebased result.
- Three stale ratchet rows naming the deleted daemon module were retired in the provider-crate policy: the rows for server state, cloud hypervisor and qemu media.
- **One caveat, recorded rather than buried:** a full-suite run during this lane showed 31 unrelated `d2bd` lib-test failures which the lane did not diagnose before its budget ended. They are outside this family's modules and need a re-run at integration to classify - the same run also carried heavy fleet starvation, so they may be that rather than anything real.
- Changelog fragment updated for the guest family.

## Review


vicondoa:
**Verdict: correct, no findings.**

**The visibility grant is exactly right**: one entry added, alphabetically placed, with a sorted set-diff confirming no other entry added, removed or reworded, no other BUILD file touched, and the scope narrowed to the single package rather than made public. It is required by three targets in the crate, and the reviewer verified it by `bazel cquery` over the library, test-support and registration dependency closures - analysis-only, zero actions, no build load.

**The newly allowed lint is a legitimate allowance rather than a suppression.** All three trait methods return a unit error, every consumer discards the payload, and every mapped outcome is a retryable unavailability or a named refusal. Decisively, the adapter this code replaced **discarded the identical underlying error**, so the unit error loses no information any caller ever consumed, and the daemon-side implementation genuinely has no richer diagnostic to carry. The allowance carries a specific reason in the workspace's convention.

**The changelog sentence matches the code**: the factory builds the service once and each created driver receives a clone of that same value, the daemon constructs one factory per zone plane, and the pinning test documents the resumed-not-restarted behaviour including the generation-fenced slots.

The two companion lint fixes are mechanical and behaviour-preserving, and nothing else moved between heads.

vicondoa:
`a81688895508ed76dffb2fb79cafd493042ed1a1`.

**Independent review: incorrect - two findings, both being fixed.**

The conversion itself is faithful, and that is the part that matters most: admission gates, budget constants, refusals and cleanup paths moved verbatim with the same error surface, composition through the generated table is single and complete, the new dependency edges are acyclic with a minimal visibility grant, and the moved tests still assert observable behaviour rather than the new seam.

Two findings:

**P2 - stale build reference.** `packages/d2bd/BUILD.bazel` still lists `src/guest_effects.rs` in the `data` attribute of the `production_binary_policy` test target (around line 472), while the same change deletes that file. A data entry naming a deleted file makes the target fail with a missing-input-file error, so `bazel build //packages/d2bd:all-tests` does not build. This was masked by a green CI run, because the daemon's Bazel test targets are excluded from the rust-main suite pending the test-support variant-graph unification - so a green run was never evidence about this target. Being fixed in the branch.

**P3 - unsupported count in the description.** The validation section claims twelve retired ratchet rows; the diff retires three (the rows naming the deleted daemon module, for server state, cloud hypervisor and qemu media). Correcting the description, because an uncounted claim in review evidence is how a wrong number survives.

---

Review bound to repository `vicondoa/d2b`, pull request #564, base branch `v3`, head

vicondoa:
**Re-verification of head `8095f48d342866ac49cfa3a921a004fa23cbcd14`: correct - both findings cleared.**

The build reference is gone, and the reviewer established something stronger than the fix: this change contains no file deletions at all, because the daemon module is a pure rename into the provider crate, so no other build or policy entry could be left naming a file that no longer exists - confirmed by a sweep of every build file at this head.

The count was checked against the diff rather than against the corrected description, which is the only way a corrected claim is worth anything: exactly three exemption rows naming the deleted module are retired, the policy file at this head carries no mention of it, and the description now states three and names them.

---

Review bound to repository `vicondoa/d2b`, pull request #564, base branch `v3`, head `8095f48d342866ac49cfa3a921a004fa23cbcd14`.

vicondoa:
**Re-verification of head `f3f6d59a749d52706d77ec90b6c98f5ceba74262`: correct - everything clear.**

The rebase was verified by presence. The incoming family arms sit beside this family arms in the driver map, the service-factory map takes both facet sets, both binding tests remain, and the visibility grant and changelog carry both. The registration table was rebuilt from the per-crate declarations rather than from either side of the conflict: four rows matching the four declarations one-for-one in identifier and service. And no hand-written composition remains for either family, which is the condition under which the generated table is genuinely the composition authority rather than a parallel path beside it.

The finding stands cleared at the new head: no build entry, no module line and no policy-list entry names the deleted module, and every file the policy list does name still exists.

The sweep was spot-checked rather than accepted: every quoted path entry across the three build files this change touches was resolved against the tree and all exist. That mattered, because a nonexistent entry of exactly that kind twice slipped past a green run today.

---

Review bound to repository `vicondoa/d2b`, pull request #564, base branch `v3`, head `f3f6d59a749d52706d77ec90b6c98f5ceba74262`.

vicondoa:
**Re-verification of head `fa9ed8054928bf249b0d7936eb1b2d1ff25eac93`: correct - everything is clear.**

The regeneration was verified line by line rather than by its file count, which is what a generated artifact needs: forty digest swaps, exactly three dependency records added per main-product closure matching the three dependencies the crate adds, and the matching entries in the lock copies - nothing else rewritten, no unrelated churn. The authority was re-run in check mode by the reviewer and exits clean without modifying the tree, so idempotency is observed rather than reported.

That closes the failure mode this branch hit: adding dependencies changes the root lockfile, and the policy inputs record a hash of it, so a three-line dependency addition can invalidate forty committed files. The fix regenerates them through the authority rather than by hand, which is the only form of that fix that stays true the next time the lockfile moves.

---

Review bound to repository `vicondoa/d2b`, pull request #564, base branch `v3`, head `fa9ed8054928bf249b0d7936eb1b2d1ff25eac93`.

vicondoa:

PR #565, opened 2026-09-21
refactor(broker,providers,d2bd): serve the systemd operations as family rows through the forward seam

vicondoa:
## What

The five systemd operations stop being broker wire variants and become family rows served through the forward seam: the wire request/response variants, their op-name and audit-join arms, the broker profile check's per-request gate and the catalog rows are removed; the supervisor's systemd effect owner moves onto the envelope carrier; and the operations are declared by the process-systemd provider crate, which now carries its operations and registration files.

Generated outputs were regenerated through the authority: broker operations, schemas, daemon API and nix inventories. `make generate` produces **no diff** against this head, so the committed artifacts match the authority byte-for-byte.

One behaviour change beyond the sweep, which is worth a reviewer's attention: the plane publishes services from driver declarations only, and the process-systemd provider has no plane driver, so its declared service was unbound. The plane now also publishes registration-declared services that no driver covers, resolving their declarations by name - and a service whose factory is missing still fails closed at startup rather than silently starting.

## Validation evidence

- **Exact `make check`** (run before the repository gate lock existed): three **deterministic** failures, none of them starvation - stale ratchet rows and new family-identifier signals the layout check caught, a drift target whose sandboxed data lacked the new operations file, and nine clippy denials in the new files. All three are fixed, each verified by its focused target: the layout check exits 0, the drift target passes, the crate's clippy and tests pass, and `xtask_test` (which includes the ratchet-consistency tests) passes.
- The refresh replayed only the lane commit: the lane's base turned out to be a local duplicate of an already-merged commit, so a plain rebase double-applied it; `rebase --onto` applied it cleanly onto the current base. The final diff is 82 files and its file list is identical to the lane commit's, so no other lane's work is reverted.
- The five rows are now family-served host-mutating operations, so they are named for the owning host-integration check rather than claimed hermetically proven; the service inventory is proven by the binding test.
- Changelog and dispositions updated; the retired rows are recorded as family-owned in the disposition document.

## Review


vicondoa:
**Independent review of head `b91751cd770d2337d55619cf7bb04ff92a33aa73`: correct, no findings.**

Covered the process-systemd family conversion and the corrected effect call sites that the same lane repaired, including the returned-response flag and the pidfd and start-failure paths.

Review bound to repository `vicondoa/d2b`, pull request #565, base branch `v3`, head `b91751cd770d2337d55619cf7bb04ff92a33aa73`.

vicondoa:
`b91751cd770d2337d55619cf7bb04ff92a33aa73`.

**Independent review: incorrect - three findings, one of them blocking.**

The move itself is a genuine substitution rather than a shim: the broker's systemd implementation is ported into the provider crate, the typed wire variants with their op-name, audit-join and catalog arms are retired with no dangling consumers or build references, the ratchet rows for the deleted module are removed, and the call-site repairs are legitimate rather than assertion-gutting.

**P1 - a refusal was deleted rather than moved, and half of it is a stale-boot replay guard.** The retired typed-wire path refused a unit operation under the Guest profile twice per request: the profile gate required the request's execution reference to target a Guest and carry a valid guest-execution binding, and the answer path separately rejected a binding whose boot-identity digest did not match the hash of the current kernel boot id. Both are gone in this change. On the new envelope path, authorisation is reduced to row membership, and the family handler's validation covers the bundle intent and the sandbox plan but never inspects the guest execution - while the supervisor still populates it and the row's payload schema still admits it as an opaque object. So a request carrying an invalid or previous-boot binding reaches the unit start, observe and stop effects instead of being refused. The description discloses removing the per-request profile gate but not the loss of the boot-identity or validity refusal. Being fixed in the branch.

**P2 - the composition root silently skips a registered service it does not name.** The new hosting pass resolves through the hand-written declaration list and continues past any registered service id that list does not spell out, so such a service is neither published nor refused at startup. The stated fail-closed posture therefore holds only for services whose declaration happens to be written down, and the registration parity gate does not catch the gap because it validates the crate's own sources rather than the daemon's list. A future conversion would compose nothing with no startup error. Being fixed.

**P3 - two comments still name the deleted systemd operations module**, in the broker runtime and its manifest, explaining a dependency that no longer exists.

---

Review bound to repository `vicondoa/d2b`, pull request #565, base branch `v3`, head

vicondoa:
**Re-verification of head `ba7916c4817b5b7b55cd8feb8b207d2ef0083dce`: correct - both blocking findings cleared.**

The guest-execution refusal is restored as both halves and is genuinely reachable. A carried binding must be structurally valid and its boot-identity digest must equal the domain-tagged hash of the freshly read kernel boot id - compared against the live boot id and never a stored digest - and a request targeting a Guest while carrying no binding is refused. All five unit handlers validate the typed request before acting, including the delete and reopen path, so no request reaches the manager leg without passing the gate; the host path is unchanged. Three regression tests pin both halves plus the matching-digest admission, with a host-target control.

The composition root now refuses startup where it previously skipped silently, on the live branch rather than behind an earlier return, and the currently registered services all resolve - so the error fires on drift and not on a valid registration.

The ancillary changes introduce no regression: the two comments no longer cite the deleted module, the test-support target links the dependencies the library target gained, chain identities are plumbed through the invocation and actor for nested legs, and the pidfd response leg now declares its fd kind in the spelling the wire contract actually accepts.

---

Review bound to repository `vicondoa/d2b`, pull request #565, base branch `v3`, head `ba7916c4817b5b7b55cd8feb8b207d2ef0083dce`.

vicondoa:
**Re-verification of head `0861358988a334d2ba657b403e592c107dac8aae`: correct - both findings and the rebase reconciliation clear.**

The rebase resolution was verified by presence: all four pieces of the incoming family survived beside this one in the composition root and the plane tests, and the generated registration table carries every converted family.

Both findings survived the rewrite: the guest-execution gate still requires a structurally valid binding whose boot-identity digest equals the domain-tagged hash of the freshly read kernel boot id, with the read live inside the validator, and still refuses a request targeting a Guest without a binding - with all five handlers reaching it; and the startup refusal for an unresolvable registration remains on the live branch.

The two things the rebase surfaced were checked against their sources rather than against the checks passing. The declaration list was extended for the family the base added, and a valid registration no longer refuses while driver-owned services still short-circuit before the resolution branch. The crate committed scope was corrected to list exactly the three privileges its declared rows carry, matched against the operation table subject and scope pairs rather than merely satisfying the authority.

Ancillary items also survived: both misleading comments are gone with no remaining references to the deleted module, the test-support target links the dependencies the library target gained, and the pidfd response leg declares its fd kind in the spelling the wire contract accepts.

---

Review bound to repository `vicondoa/d2b`, pull request #565, base branch `v3`, head `0861358988a334d2ba657b403e592c107dac8aae`.

vicondoa:
**Re-verification of head `b32c1f28d0e46978dd894393a1a121c4354c67c2`: correct - delta verified, prior findings remain cleared.**

The delta is one file and two lines: the initializer that the base delivered, in a crate this branch does not own, now names both fields the branch made required on the shared invocation type. Nothing else moved, so the guest-execution refusal with both halves, the startup refusal for an unresolvable registration, and the rebase reconciliation all remain as cleared.

The tree-wide claim was spot-checked rather than accepted, which is what made it worth asserting: every initializer of the invocation type and of its daemon-side carrier was checked individually, and each names both fields - including the one that spells the field by shorthand, which is valid rather than a miss. That matters because this defect existed precisely because the failing site was outside the branch s own paths, and only the per-target Bazel compile saw it.

---

Review bound to repository `vicondoa/d2b`, pull request #565, base branch `v3`, head `b32c1f28d0e46978dd894393a1a121c4354c67c2`.

PR #567, opened 2026-09-21
refactor(activation,d2bd): serve the activation family from its own crate

vicondoa:
## What

The activation family's effects move out of the daemon: `packages/d2bd/src/activation_effects.rs` is deleted, the crate declares an effects service (`activation.d2bus.org/effects`, method `inspect-activation`), its facets, a registration and a recording double in the release-gated test support, and the daemon composes the family through the generated registration table with the hand-written family arm removed. The daemon retains only the broker dispatch its composition root supplies, wrapped in the crate's facet trait.

The response reduction the family performs is preserved, and three stale ratchet rows naming the deleted module were retired. Generated artifacts came through the authority: `make generate` exits 0 and produces no further diff.

## Validation evidence

- **Exact `make check`: 922 of 923 passed.** The single failure is the known broker load flake, which passed on a re-run in isolation. The layout authority passes, and so do every drift gate - including the policy-inputs and nix-inventory ones, which matters here because the family gained a dependency edge.
- Focused evidence: the crate's own suite green through its test-support feature (six new effects-service tests plus driver, controller, registration and reconcile), and the plane's lib tests green including the two new hosted-service and restart-adoption tests.
- Changelog fragment added.

## Review


vicondoa:
`c5dac4ae5aec55e386e15e0258014413de874ce9`.
**Independent review: correct, one finding.**

The conversion is sound. The activation family's effects genuinely moved into the provider crate - the daemon retains only the declared facet supply and composition, and nothing is re-exported or shimmed back. The rebase resolution is complete: both sides kept in the plane's conflict blocks (host and activation facets, both factory arms, both driver arms, both test groups), and the registration table and policy inputs regenerated from the authority rather than hand-resolved.

One finding, being addressed before merge:

**Capability width (P3).** `ActivationBrokerDispatch` is declared as `fn dispatch(&self, request: BrokerRequest) -> Result<BrokerResponse, String>`, and the daemon-side implementation dispatches whatever request the crate hands it under the daemon's admin caller authority. The family needs exactly one operation - the preserved host-generation handoff - and the sibling families already converted declare narrower domain facets (`NetworkRuntime` for the network-local family; a provider-runtime trait plus identity-source traits for process). A facet shaped to the one preserved operation keeps the crate to its committed vocabulary instead of authorising every broker request variant; the domain-facet shape its siblings use is the existing pattern here.

---

Review bound to repository `vicondoa/d2b`, pull request #567, base branch `v3`, head

PR #569, opened 2026-09-21
docs(broker): record why these file modes are deliberately unpinned

vicondoa:
## What

Two comments, nothing else - eleven added lines in the broker's network and state-cell modules, at the opens that request a group-readable mode without pinning it.

They exist because of a hazard another change created: the export channel and the audit path are now *pinned* to an exact mode, for a real reason - their readers depend on group-read through a default ACL, and a restrictive service umask was silently stripping it. A reader who sees two group-readable opens pinned will reasonably grep for the rest and pin those too, and at these two sites that would be **wrong**: the rows are broker-root-only, no declared consumer needs the group bit, and a post-open `fchmod` would *widen* the file relative to what the operator's umask chose. A umask can only remove bits, so the failure direction here is widening, not leaking.

## Why a comment rather than a code change

Six opens were investigated for the same suspicion and **none** warranted a pin: the durable writer of the third already pins through its atomic replace, and the state root's ownership and ACL mean the group bit was never a working channel for a non-root consumer. So there is no defect to fix - but there is a plausible wrong fix to prevent, and one comment at each of the two sites the durable-row family mirrors is the cheapest thing that prevents it.

## Validation evidence

- Focused run of the two touched modules: 21 passed, 0 failed.
- Comment-only diff; no code line, mode or assertion changed.
- The investigation record - six sites, the reasoning per site, and two adjacent directory-creation modes with root-only consumers - lives with the tracking item rather than in the tree.

## Review


vicondoa:
**Verdict: correct, no findings.** The comments' claims were verified by tracing every reader of both files (all broker-internal), the two pin sites the advice refers to, the production umask, and the directory/ACL posture. Three residuals are recorded here because they belong with the PR rather than in another head change:

1. **The shipped service umask is `0027`** (both broker units), not an arbitrary one. Under it an open-time `0640` stays `0640`, so a pin at these two sites would be a **no-op rather than a widening**. The unpinned choice is correct either way - no consumer needs the group bit - but the comment phrases the umask point generally, and a reader could take it as describing production as `0600`. Worth knowing when reading it.

2. **Do not generalise the advice to the nft durable writer.** Its pin *is* load-bearing: its consumers include the unprivileged CLI and the status reporter, which need read on a file a restrictive umask could narrow. The comment is scoped to these two paths and does not extend there, so nothing needs correcting - only an over-generalising reader could misapply it.

3. **The ACL caveat.** The row directories are `0750` root-owned and the state root is traverse-only for the daemon group with no default ACL, so the rows are effectively root-only today despite being created `0640`. If an operator ever adds a default ACL or a group grant to those directories, the unpinned mode would make the rows group-readable - benign for this data, and worth re-checking then rather than now.

vicondoa:

PR #571, opened 2026-09-21
test(broker): derive the network test scratch root from the helper

vicondoa:
## What

Three network realization tests built their scratch root as `current_dir()/target/network-realization-<pid>` and created it with a **leaf-only** directory builder. Cargo runs test binaries with the package root as the working directory, so they required `packages/d2b-broker/target` to already exist - and in a fresh checkout it does not, because the workspace target directory lives at the repository root. The result: all three failed **by construction**, in about 0.06 seconds, in every clean worktree.

This had a real cost. Two separate lanes hit those three failures, re-ran them under different conditions, and classified them as load flakiness - which is a reasonable inference from three tests failing together on a loaded machine, and wrong.

They now derive the root from the crate's shared test helper, which creates its base directory and does not depend on the caller's working directory. The remove-then-create semantics, the file mode, and the registered lint allowances are unchanged, and no assertion was touched.

## Validation evidence

- **The acceptance criterion is the condition the old code failed**: with `packages/d2b-broker/target` **absent** before the run, `cargo test -p d2b-broker --lib realization` -> **3 passed, 0 failed**, in 0.03s. The directory is still absent afterwards, so no workaround was created or left behind.
- The whole module passes: 9 tests, 0 failed.
- **Class sweep**, since I would rather know the size than rediscover it one lane at a time: three sites were broken by construction (fixed here); twelve others share the same working-directory-relative smell but survive a clean checkout because they create their parents, and are candidates for the same migration as a follow-up; three are deliberate working-directory-relative *lookups* of built binaries and tools, which is correct; three are not this class at all. The follow-up candidates are listed in the tracking item rather than in the tree.
- Changelog fragment added.

## Review


vicondoa:
**Verdict: correct, no findings.**

Verified: the helper never touches `current_dir()` - it resolves through the test temp variables and creates its base with a single expect, so it cannot fail in a fresh checkout short of an unwritable temp base. The remove-then-create sequence and the `0o750` mode are unchanged, and the code under test validates only the state directory's own mode and the file modes it creates, so there is **no parent-mode dependence and no silent behavioural change**. The diff touches only root-derivation lines: assertions, fixtures and scenarios are identical, and the focused run passes with no leftover directory.

The changelog's framing was checked and is accurate: cargo runs unit tests with the package root as the working directory and the directory builder is non-recursive, so the tests only ever passed in checkouts where another test had already created the parent.

**And the follow-up judgement, which matters more than the fix**: two of the twelve sweep survivors were spot-checked and both create their own parents, so they pass in a clean checkout. Their residual problem is a package-relative scratch *leak* - hygiene, not correctness. That makes the migration a cleanup lane rather than a fix-for-failure lane, and it should be prioritised accordingly rather than as another reliability task.

PR #572, opened 2026-09-21
refactor(interaction,d2bd): serve the interaction and desktop family from its crates

vicondoa:
## What

The interaction and desktop family's implementation moves out of the daemon into its crates: the interaction effects, the audio registry and the child-source implementations now live behind declared facets and services, the daemon keeps only the facet implementations its composition root supplies, and the family starts through the generated registration table. Three daemon modules are deleted.

This lane had never been compiled, so its first build surfaced the mechanical work of finishing the design: the audio mediator needed the thread-safety bounds its facet shape requires (and a boxed-value delegation so the boxed trait object its registry holds is still a mediator), a stale module declaration and two unused imports went, a never-read facet field went, and three test files repointed imports at constants that had moved crates.

The layout authority demanded six exemption rows, and each was argued rather than assumed: two for the shared crate naming the audio family's crate behind the facet the daemon hosts (identical in kind to the rows its audio dispatch already carries), and four module-and-token rows for the family that *composes* sibling families - its vocabulary must name their declared type ids. All six are module-plus-token specific; none is a blanket relaxation.

## Validation evidence

- **Exact `make check`: exit 0, 923 of 923 tests pass**, under the repository gate lock with the resource cap and no profile override.
- Three earlier gate attempts produced no result: one was starved (mass NO STATUS, treated as starvation rather than failure per the standing rule), one exposed a real analysis error, and one exposed a real clippy denial. Both real errors were fixed - **three crate visibility grants** and a redundant closure - and the fourth run is the green one.
- The lane's own crates: `cargo check --all-targets` clean and their Bazel package suites pass 65/65 including clippy.
- The layout authority is idempotent at this head (writes 14 artifacts, none drifted), and `make generate` changed only policy inputs - exactly the new dependency edges.
- Changelog fragment updated.

## One thing to flag for review

The family's *effects reconcile sibling families' declared resource types*, which is why its vocabulary names their type ids. If that architecture is wrong, the four exemption rows are the symptom rather than the fix - worth a reviewer's judgement rather than mine.

## Review


vicondoa:
**Independent review of head `0fa7ffe1afd90bbd450f19453b968be4d2c8271a`: correct, no findings.**

Covered the interaction and desktop family conversion, including the projection-dependent reads and the composition wiring the conversion rebuilt.

Review bound to repository `vicondoa/d2b`, pull request #572, base branch `v3`, head `0fa7ffe1afd90bbd450f19453b968be4d2c8271a`.

PR #573, opened 2026-09-21
refactor(device,d2bd): serve the device families from their crates

vicondoa:
## What

The device families' effects move out of the daemon into the five device crates. Each declares its own facets, effects service and registration where it has resource types, and the daemon composes them through the generated registration table, keeping only the facet implementations its composition root supplies. The daemon module that held the TPM effect port is deleted, the USBIP dispatcher and its ledger move behind a declared facet so the fail-closed attach seams travel with the code that owns them, and the per-component device orchestration stays daemon-side behind the runtime trait.

This lane was never compiled before its continuation, so most of its work was mechanical completion: stale lockfile entries, a library/dev-dependency misplacement, unused imports, a boxed-borrow lifetime, a lock type that did not match the port's declared contract, and a test that asserted against the wrong double once the driver began routing through facets.

The layout authority demanded four family-knowledge rows and retired twelve stale ones; the parity fix also required regenerating the committed registration table, which the check caught as drift rather than letting a hand edit pass.

## Validation evidence

- **Layout authority: PASSED** - and only after the check's own fixer was run, because the parity fix had never regenerated the committed table. Its first objection was exactly that drift, which is the check doing its job.
- **Exact `make check`: exit 0, 924 of 924 tests pass**, under the repository gate lock, with no NO STATUS targets - unlike this lane's earlier runs, which were starved. The first attempt failed one target because the invocation lacked the shellcheck tool the Nix shell exports; re-run with the declared tool, it is green.
- **Per-crate Bazel verification: 73/73** across the five provider crates including every clippy variant; the daemon's own package targets pass except one clippy target that hits the documented worktree build artifact (duplicate crate identity in a pre-existing test-support closure).
- Changelog updated.

## Notes for review

Two structural decisions are worth a reviewer's judgement rather than mine, and both are recorded in detail: the five static-service factories carry a facet field for the respawn contract that their  does not yet read (allowed with a reason, uniform across all five), and the daemon-side construction-input facet fields for TPM and GPU were removed as never read because the facets flow through the attach path instead.

## Review


vicondoa:
**Verdict: correct to merge** - every refusal, admission, ownership-marker and fail-closed path survived the move, quoted where each does. In particular: the GPU row-owner fence and admission owner checks are byte-for-byte the retired port; the USBIP privileged dispatch stays daemon-hosted with its fail-closed mapping and the same admin-uid authority; the USBIP and TPM/GPU facet seams refuse retryably rather than silently no-op; TPM caller authority and the integrity refusals are unchanged; and a declared service with no factory still refuses at startup. The refined error mapping flips nothing in practice - the three retryable arms are unreachable, so every real failure stays terminal - and the lock-type change removes only a poisoning path that no user code could observe.

**One item to fix before merge:** the respawn-contract comment in all five factories asserts that `build()` rebuilds from the facet field, and `build()` constructs a static service and reads nothing. That is a comment contradicting its code, in the exact place the respawn-contract implementer will read it. It is being corrected rather than recorded.

**Two residuals, recorded not fixed:** the provider startup order changed, with the test re-pinned but the consequences of the earlier position not separately analysed; and the facet attach silently ignores a rejected double-set, which is harmless at its single call-site but would hide a second attach.

vicondoa:
**Comment correction applied, and two questions answered from the code.**

The five factory comments now state what the code does: the facet set is carried for the respawn path, that path is **not** wired, and the current static service does not read it. The author chose to record the contract honestly rather than implement the rebuild now, on the grounds that the respawn lane is separate work and wiring it would give the static inspect services state they do not need - I agree with that call.

**Startup order: functionally inert, and grounded rather than asserted.** Each provider's attach is a self-registration only; envelopes are built from each provider's own drivers; every order-sensitive check is global and runs before or after the whole attach pass rather than interleaved; drivers come from construction inputs with no cross-family reads; and drain mirrors startup as a per-provider self-release. The only consequences of the new order are the pinned sequence and the durable-row insertion order, both deterministic.

**Double attach: impossible today, but fail-open by shape.** One call site, crate-private, and tests construct the input struct directly rather than attaching - so the ignored rejection is unreachable. But if a second attach path is ever added, the first facet set would silently win and the second would be dropped, which is exactly the class this program refuses. Recorded as a latent trap rather than fixed, since changing it now would be speculative.

vicondoa:
**Verification of the corrected comments: correct, no findings.** The delta is comment lines only - a whitespace-ignoring diff shows zero code lines - and the static service construction is unchanged in all five factories. Two residuals recorded: the comments will need revisiting when the respawn path is implemented (nothing enforces that, which is inherent to comments), and the daemon-registration claim could drift if the composition-root wiring moves.

vicondoa:
**Re-verification of head `277b55f8f21245fc988f82c706b4dd323268922f`: correct - all clear.**

The restored coverage genuinely covers the gap rather than naming it: the test calls the moved row-template function directly and asserts both video postures launchable through the video role, so dropping either posture makes the row refuse and fails the assertion. It also rejects a posture presented under the wrong role and refuses an undecodable row instead of reading it as a posture - and it exercises the same gate the launch path consults, which is what makes it coverage rather than decoration.

The attribute-attrition sweep was re-run independently at this head and finds zero base-attributed tests surviving without an attribute. The only two attributed tests this conversion deleted were both GPU ones: one moved to the crate with its attribute intact, and the other is the one restored here.

On the distinction that matters for a judgement like this - a deleted surface versus lost coverage - every deleted line in the diff was searched for the effects-service vocabulary and nothing matched. What disappeared is the daemon-built port surface the conversion deliberately retires, with the reconciliation bodies renamed onto the runtime trait impls unchanged. A surface deletion, not coverage orphaned by a move.

---

Review bound to repository `vicondoa/d2b`, pull request #573, base branch `v3`, head `277b55f8f21245fc988f82c706b4dd323268922f`.

vicondoa:

PR #574, opened 2026-09-21
test(broker): stop measuring wall time in budget and reaper tests

vicondoa:
## What

The broker's budget, activation and reaper tests no longer measure wall time or sleep before asserting. Each site was converted by failure class rather than mechanically:

- **Virtual time** where the code under test runs on the test's own runtime: the round-trip budget test now pauses the clock and asserts the refusal falls inside a one-millisecond virtual window, with the peer provably still holding the connection so the refusal cannot be the peer's close; the connect-refusal test asserts zero elapsed on a paused clock.
- **Causality where virtual time is impossible, with the reason recorded**: the spinning-handler abort test's handler runs on the global real-time worker set, so a paused clock would auto-advance past the probe - the author verified that by running it. It now takes a progress-based started signal from inside the spinning handler and relies on the refusal code, which only the budget-abort branch produces.
- **Progress-based readiness** instead of a fixed sleep before connecting: the activation test retries the whole handshake, and the shared helper probes the *actual* readiness condition - a successful connect - rather than the socket file's existence. File existence is not readiness, and probing it turned a confusing connection-refused failure into a fast, well-explained one; that change also fixed a guest-profile flake as a side effect.
- **Event-driven waiting** for the reaper tests: a notification is signalled on every reap buffer push, so the waits are event-driven with a generous guard rather than deadline loops, and a racy start-time read was removed - a child that exits before the read reports no start time, so the test could observe a live child that was already gone. The reservation path never consults that value.

## Validation evidence

- Every owning target passes ten consecutive runs under ambient load, and the reaper group passes in isolation.
- **Exact `make check`: 923 of 923 pass**, run plain with the resource cap and no profile override; the earlier failures in the same window were starvation and the then-unfixed sibling sites, both since addressed.
- The activation readiness change was demonstrated by its effect on a flaky sibling test (2/5 failing before, 5/5 after).
- Changelog fragment added.

## Review


vicondoa:
**Independent review of head `186ed2c6a63ca39bdfec3199a18618808d63ae7c`: correct, no findings.**

Covered the broker budget, activation and reaper tests, which no longer measure wall time and instead assert the observable outcome of each budget.

Review bound to repository `vicondoa/d2b`, pull request #574, base branch `v3`, head `186ed2c6a63ca39bdfec3199a18618808d63ae7c`.

vicondoa:

PR #575, opened 2026-09-21
refactor(endpoint,binding,d2bd): serve both families from their own crates

vicondoa:
## What

Two families move out of the daemon at once, because their effect implementations shared the same two daemon modules. The endpoint and volume-binding crates now declare their own facets, effects services and registrations; the daemon keeps only what genuinely belongs to it - the socket-target registry, the plane-table evidence probes and the zone's target directory - all crossing as declared facets, and the composition root starts both families through the generated registration table with no hand-written family block remaining.

The purpose derivations move with the family that owns the vocabulary, gaining dependencies on the two sibling crates whose constants they read - no cycle, and the visibility grants those edges need are part of this change.

One judgement call worth a reviewer's attention: a handful of shared-ratchet rows were retired and **no** provider-exemption rows were added, because the family scanner skips the crate-reference form the derivations use. The layout check passes, which is the evidence, but an exemption-free outcome deserves a second look rather than my assurance.

## Validation evidence

- **Exact `make check` under the repository gate lock: 921 of 923 pass.** The two failures are the known broker load flakes, each passing on an identical isolated re-run - not this lane.
- Focused: both provider crates' `all-tests` pass, the daemon binary builds, the plane's lib tests pass including the startup-order test with the new declaration order and the composition-root hosting test, the layout authority passes after `--fix` and on a plain re-run, and the generator ran clean.
- No assertion was weakened. The deleted daemon modules' own test blocks moved with their subjects into the crates' service tests.
- Refreshed twice onto the current base; the final diff is 70 files, all this lane's paths, with no other lane's work reverted.
- Changelog updated.

## Known limitation

No plane-level test resolves `inspect-endpoint` or `inspect-binding` end to end: the generic hosting test plus the crates' own service tests cover the surface, and a plane-level test would be a follow-up rather than a gap in this change.

## Review


vicondoa:
**Re-review of head `260e43db37df760ba8e8447d39683b7daefb9eb4`: correct - mergeable.**

All three items verified at this head. The two new plane tests start the plane, resolve `ENDPOINT_EFFECTS_SERVICE` and `BINDING_EFFECTS_SERVICE` through the plane, call the hosted method with real call data, and assert the inspect payloads - the previously untested hosted path, and they fail if hosting, resolution or the response drifts. The socket purpose refusal is genuinely restored: it is the first statement of present, ensure and remove, checked against the family crate purpose token before any registry access, with the message the pre-move effect carried, and the delete path routes every non-evidence purpose through it so it fires there too. The burst test waits on the projection rather than a wall-clock window and asserts fewer re-materialisations than notices, so an un-coalesced drain still fails. Every finding from the previous verdict is cleared.

The conversion substance was a pass in the earlier verdict: the effects are implemented inside their owning crates with the daemon modules deleted and no surviving references, composition runs solely through the generated table, the new edges are acyclic, the visibility grants are minimal, and the ratchet changes tighten rather than widen.

---

Review bound to repository `vicondoa/d2b`, pull request #575, base branch `v3`, head `260e43db37df760ba8e8447d39683b7daefb9eb4`.

vicondoa:
**Re-verification of head `9be4bd21009205cb31f2cc8273d76eaa8a41f132`: correct - mergeable, previous findings remain cleared.**

The rebase resolution was verified by presence rather than by compilation, which is the only way that operation can be trusted: the construction inputs carry the incoming family alongside this one in both the production root and the test rig, the service-factory signature and body handle all five services, the driver arms keep the incoming arm beside both of this branch s, the test factory map carries all three entries, the changelog keeps every bullet, and the committed startup order matches the regenerated registration table s declaration order exactly. The two daemon modules stay deleted with no shim references and no leftover ratchet exemptions.

On the amended commit: no conflict marker exists in any pushed tree - a tree-wide scan at this head finds none, and every conflict site retains the union of both sides - so the marker lived only in the lane s unpushed state and never reached the pull request.

The three earlier fixes are present unchanged: the family-owned purpose token read by the daemon facet, the restored refusal as the first statement of present, ensure and remove including the delete path, both hosted-method plane tests resolving the declared services through the plane with real call data, and the burst test waiting on the projection and asserting fewer re-materialisations than notices.

---

Review bound to repository `vicondoa/d2b`, pull request #575, base branch `v3`, head `9be4bd21009205cb31f2cc8273d76eaa8a41f132`.

vicondoa:

PR #576, opened 2026-09-21
refactor(credential,d2bd): serve the credential family effects from its own crate

vicondoa:
The credential family effects move out of the daemon into the provider crate that owns them, behind declared facets, with the daemon reduced to facet supply and composition through the generated registration table.

The daemon module that implemented the family port is deleted rather than re-exported, so the crate that declares the port is the crate that serves it.


vicondoa:
**Independent review of head `3ae65894eba5332a409be0b2f5fda2ca2b727c0d`: correct.**

The conversion is genuine and the substitution question was answered explicitly rather than assumed. The daemon module that implemented the family driver effects is deleted with no re-export, and the crate now constructs its own effects service from the declared facets, with the daemon supplying only the runtime facet it genuinely holds - the store rows and the supervisor registry the crate cannot own. The hosted service declaration rides the descriptor and resolves through the generated registration table, and a declared service without a hosting factory refuses startup by name, so no hand-written composition remains beside the table and the family cannot be left unresolvable or composed twice. Every pre-move refusal survives: the lease-absent skip, the fail-closed dependency reads, and the unconditional session handoff. The deleted modules leave no stale references in source, build files or policy files.

Two informational notes are recorded rather than treated as defects, and the reasoning is right: the daemon-implemented runtime facet is still the deciding implementation over the daemon-owned state, which matches the pattern the already-landed runtime facets set, while the classification logic over those facts lives in the crate.

---

Review bound to repository `vicondoa/d2b`, pull request #576, base branch `v3`, head `3ae65894eba5332a409be0b2f5fda2ca2b727c0d`.

vicondoa:
**Re-verification of head `f674f0120ccff26ded7ff4d6461ac1a2b9967c61`: correct - everything is clear.**

The rebase was verified by presence. The registration table carries every converted family and matches each per-crate declaration byte-for-byte under the drift-gate header, which is the only form of that check worth having - it compares the generated artifact against its authority rather than against a hand resolution. The composition root carries every family together with each descriptor constructor appearing exactly once, and no hand-written block remains beside the table, so the table is genuinely the composition authority rather than a parallel path.

The conversion substance survived untouched: the diff across this family crate sources between the previous head and this one is empty, the daemon modules remain deleted with no re-export or reference in source, build files or manifests, and the startup refusal for an unresolvable registration is unchanged. The startup-order test reflects the table order.

---

Review bound to repository `vicondoa/d2b`, pull request #576, base branch `v3`, head `f674f0120ccff26ded7ff4d6461ac1a2b9967c61`.

vicondoa:

PR #577, opened 2026-09-21
refactor(volume,d2bd): serve the volume family from its own crate

vicondoa:
The volume family effects move out of the daemon into the provider crate that owns them: the adapter and its suite relocate with the behaviour, the crate declares its facets, effects service and registration, and the daemon composes the family through the generated registration table with the hand-written block removed.

The declared but unbuilt neutral contract and its host wrapper retire with the family, as the plan scheduled rather than as collateral. The moved suite needs the network-local dev-dependency the Bazel graph already carried, and the crate declares its test registration the way the other converted crates do.

The daemon module that implemented the family port is deleted rather than re-exported, so the crate that declares the port is the crate that serves it.


vicondoa:
**Independent review of head `4179c701930847b41b668550c57c2702bc399173`: correct - three small findings, being closed.**

The conversion genuinely substitutes rather than shimming: the 2,024-line daemon adapter moved into the provider crate at 97% similarity with only doc and import-path changes plus one erasure implementation, the daemon s hand-written volume composition is deleted with no re-export surviving, and the plane now composes the family through the generated registration row, with the provider-owned service delegating only to the daemon-supplied facet.

The rebase resolution was verified rather than assumed, and it is worth recording that it initially dropped host-side syntax - a cast in the service-factory map and three closing parens in the plane tests - which the final commit restores; both families sides are present at head, as are both sides of the visibility lists.

Both regenerated artifact sets were verified by re-running their authorities rather than by reading them: the layout check confirms the registration table is byte-faithful to the per-crate declarations, and the policy-input generator confirms all 76 inputs match what the authority computes. Deleted-file references were checked by hand across build files, manifests, ratchet tables and policy JSON with nothing stale, and the three ratchet exemptions naming the removed modules are removed in the same change. Both declared deletions are scheduled retirements with their ledger, dossier and changelog entries.

Findings, all P3 and all being closed: the registration test does not pin the declared effects service, which the sibling host lane does pin, so a regression dropping it from the descriptor would pass this suite while removing a startup refusal; a test-helper comment claims the runtime double refuses every call when it records and succeeds; and a ledger entry still records the retired contract as outstanding work.

---

Review bound to repository `vicondoa/d2b`, pull request #577, base branch `v3`, head `4179c701930847b41b668550c57c2702bc399173`.

vicondoa:
**Final re-binding of head `cf3a3dffd911ec0af6a17a286baeb73af441a751`: correct - all findings clear.**

The repaired ledger text was verified for meaning rather than for tidiness, which is the distinction that matters: it now states the retirement accurately against the review facts, and the ownership cell converts the original flawed claim into a correct one rather than trading one wrong statement for another.

The removed declaration constructor left nothing behind - it has no matches anywhere in the tree including test code, the imports it kept alive have no remaining uses, and the build is clean of unused-import warnings. It also confirmed the function died by the intended cutover rather than by an accidental caller removal: its only caller was the daemon s hand-written composition block that this conversion replaced with the registration table, and the daemon s own family declaration produces the identical declaration.

The non-ASCII scan covers the whole delta, added and removed lines alike, with no matches.

---

Review bound to repository `vicondoa/d2b`, pull request #577, base branch `v3`, head `cf3a3dffd911ec0af6a17a286baeb73af441a751`.

vicondoa:
**Re-verification of head `a5f95df270fe3d5e0abcd17738f62c914854079c`: correct - everything clear.**

The rebase was verified by presence. The composition root carries every converted family arms together and the registration table carries every row, with the layout authority drift gate exiting clean at the new head - so the table was regenerated from the per-crate declarations rather than carried from either side of a conflict, and no row was lost. That is the check that matters for a rebase, because a lost table row compiles perfectly and fails only at startup.

All three earlier findings survived unchanged: the assertion pinning the declared effects service, the refusing double wired into the registration boundary, and the dead declaration constructor with zero remaining references. The repaired ledger prose is intact and the delta adds no non-ASCII characters.

---

Review bound to repository `vicondoa/d2b`, pull request #577, base branch `v3`, head `a5f95df270fe3d5e0abcd17738f62c914854079c`.

vicondoa:

PR #578, opened 2026-09-21
test(broker,d2b,ci): measure behaviour rather than the machine, and survive a bad download minute

vicondoa:
Four gate-reliability changes land together, because they are one unit: each exists so that a test reports the code rather than the machine it ran on, and each was waiting on the same pipeline.

The broker budget, activation and reaper tests now assert the observable outcome of each budget instead of how long it took. The CLI, bus, supervisor-adapter and guest-controller timing barriers are rewritten the same way. The two comments that recorded deliberately unpinned expectations are corrected rather than left asserting a pin that is not there. And CI now retries the Bazel fetch and analysis phase when a third-party release download returns a transient server error - bounded, triggered only by that rendering, with real failures still failing on the first attempt and a persistent outage still failing after the bound.

Combining them was not merely a queue reduction: it surfaced a cross-change regression that none of them would have shown alone. The workflow retry moves the make invocations behind an anchor and an environment variable, and the test merged separately that counts literal make invocations in the workflow could no longer see them, so that pull request alone would have failed CI against the current base. The test now counts both spellings at the same threshold, and the intent it exists for - that CI exercises the public make aliases directly rather than through nested nix wrappers - is preserved rather than relaxed.


vicondoa:
**Independent review of head `81eadf60491daf144f90736b4af1fab00d6089b4`: correct - mergeable, no findings.**

No constituent was dropped in the squashing: all four are present at this head, verified by inspecting their own files rather than by the change compiling. The conflict resolution is sound rather than a relaxation - the counting test now recognises both the literal and the anchored spelling at an unchanged threshold of ten, the workflow still contains exactly eleven make invocations against eleven at the base, and the retry runs the public alias verbatim rather than routing around it, so the property the test exists for is still enforced and the assertion guarding it is untouched.

The retry is genuinely fail-closed: an exit whose log lacks the transient-server-error rendering fails on the first attempt with its real status, and a persistent outage fails after the bound with the last status, with the real exit status captured through the pipeline under strict shell mode.

Every rewritten timing test still has a condition under which it fails, which is the property that distinguishes an outcome assertion from a test that can no longer fail: the budget tests assert the refusal mechanism or an exact virtual-time window whose in-loop assertion panics when the budget stops binding, the dial test asserts zero virtual-time consumption, and the readiness and reaper waits assert buffer contents and exit kinds against documented last-resort guards whose magnitudes were widened deliberately and are commented at the change site.

One coupling is documented in the code and was verified against it rather than accepted: a session wait is sound only because the keepalive interval exceeds its guard.

---

Review bound to repository `vicondoa/d2b`, pull request #578, base branch `v3`, head `81eadf60491daf144f90736b4af1fab00d6089b4`.

PR #579, opened 2026-09-21
feat(xtask): check that a declared dependency edge has a bazel visibility grant

vicondoa:
A provider crate that declares a dependency on a sibling package compiles only if that package grants it visibility, and the grant is easy to forget because the failure surfaces as a Bazel analysis error in whichever target is built first - which, given that the daemon test targets are excluded from the main suite, may be nobody in CI.

The layout authority now resolves each declared dependency edge against the visibility grant, package-default or target-level, treating genuinely public targets and public re-export aliases as granted structurally rather than requiring exemptions. A missing grant is reported with the consuming crate, the owning package, the rule and target, and the exact entry that is missing.

The gate was demonstrated failing by removing one grant and restored byte-identically afterwards. Its non-coverage is documented rather than implied: transitive rather than declared dependencies, data and tools edges, target existence, and consumers outside the provider class.


vicondoa:
**Re-verification of head `12e79e093b0920b63f3b6be21f3bc763741160bf`: correct - the finding is cleared.**

`proc_macro_deps` is genuinely gated rather than merely parsed: both attributes flow through the identical grant-resolution loop with no separate path, and the new fixture exercises both the refusal and the target-level grant that satisfies it. The attribute lookup is now exact, and the fix addresses the suffix-embedding accident in general rather than only the reported instance - the same mistake applied to a visibility lookup inside a default-visibility attribute - with the skip provably non-lossy, since it advances only past the needle itself, so a genuine attribute later in the same body is still found. The documentation now names the edge kinds it covers and the ones it excludes, so a reader cannot conclude broader coverage than exists, which was the substance of the finding.

The independent model was re-run over the committed provider build files at this head: sixty files, one thousand and seventy-one gate-relevant cross-package edges, zero false positives and zero false negatives. The grant logic, consumer set and diagnostics are otherwise byte-identical to what was verified in the previous round.

---

Review bound to repository `vicondoa/d2b`, pull request #579, base branch `v3`, head `12e79e093b0920b63f3b6be21f3bc763741160bf`.

PR #580, opened 2026-09-21
fix(d2bd): read the row identity rather than a metadata projection

vicondoa:
Several checks read metadata out of a JSON projection where the authoritative value is the row own identity, and the projections differ per call site: one carries raw stored metadata, another the rendered envelope. That is why some of these checks appeared to work - the field happened to be present - and others could never pass.

The credential runtime compared the request zone against a projected zone. The gateway custody check compared a row-derived value against itself, so it could never fail for a correctly resolved row and failed deterministically when the projection did not synthesise the field. The shared provider effects read uid and generation from the projection rather than from the identity the manager holds for that row.

The network admission zone fence is a different case and was not removed: its request zone is external input, so the comparison is precisely what keeps a cross-zone request from admitting rows resolved under this plane. It now compares the two zones actually in the call - the zone the rows were resolved under against the zone the request names - and removal was rejected because it would have opened that boundary. A same-zone attached Guest is admitted end to end, proven in a clean network namespace since the host environment refuses on unrelated real-route grounds.


vicondoa:
**Independent review of head `c14269efa094d9572a9ac1be68b618455943bab3`: correct - no findings.**

The three kinds of change were verified separately, which is what this diff needed.

The credential runtime redirect sits entirely in test-only code with no constructor anywhere and no test exercising the branch, so it has no reachable effect and weakens nothing.

The removed gateway custody comparison was checked on both halves of its claim and both hold. The read key is pinned to the plane own zone, so a projected zone could only ever equal it - the comparison was a tautology that could not fail for a correctly resolved row. And Nix-materialised rows carry no zone in their raw stored metadata, so it refused deterministically on every correct request while never discriminating a wrong one. Paths that do author a zone validate it against the plane own, so no boundary acting on a different input was removed, and the surrounding refusal and unavailable paths remain.

The identity reads now take the manager-authoritative top-level values the response actually carries, which the raw metadata does not.

And the admission fence is equivalent in strength to the one it replaces for every input the old comparison covered: all rows in both loops are homed in the effects own zone, so comparing that zone against the request zone is equivalent to the old row-zone comparison, a cross-zone request is refused exactly as before, and no path into admission bypasses it. That equivalence is the point - the fence was redirected rather than removed because the request zone is external input.

---

Review bound to repository `vicondoa/d2b`, pull request #580, base branch `v3`, head `c14269efa094d9572a9ac1be68b618455943bab3`.

vicondoa:
**Re-verification of head `ce2602590a392488a5c7f581d6a67b05ea0f950b`: correct - everything is clear.**

The security fix moved exactly one test assertion and left the five functional fixes byte-identical. The new message has no source from which an identifier could reach it: the formatted value is either a literal or the display form of a fieldless error enum whose display writes one of three fixed tokens, none of which is a name, uid, path, principal or credential.

The diagnostic value is preserved rather than traded away, which is the property that makes this a fix and not a suppression: those three tokens correspond one-to-one with the refusal classes, so a failure still identifies which refusal occurred instead of becoming a generic message that hides it.

The rest of the change was spot-checked against the diff rather than accepted: it adds no logging or printing statement at all, and every other interpolation is fixture construction - a resource reference name from the test own constant, and storage-row identifiers synthesized from the fixture zone - none of which is rendered into a log, a panic or an assertion message. The one instance the scan found is the one this commit fixes.

---

Review bound to repository `vicondoa/d2b`, pull request #580, base branch `v3`, head `ce2602590a392488a5c7f581d6a67b05ea0f950b`.

vicondoa:
**Re-verification of head `fcfeb791deb014c1612f1d1b336f574a0638a7c3`: correct - everything is clear.**

The scaffolding fix adopts the shapes the base now uses rather than deleting what it cannot name: the two fields are renamed to the facts structures the base introduced and built with the same doubles the base own test inputs use, and the factory map gains the entries matching the composition helper entry for entry, including the one family whose factory carries no facet set by design. The old port fields no longer exist on the base type, so updating was the only option - no coverage was removed, and the only surviving pre-conversion double belongs to a family whose port field the base still carries.

The validation reasoning is the part worth keeping: the failure was a compile error against field names the base removed, so it can only manifest when the branch tree includes that base - at the merge ref, and never on the branch own pre-rebase tree. That is a complete explanation for a failure that looked transient for hours, and it is the same mismatch that made several of today failures hard to read.

---

Review bound to repository `vicondoa/d2b`, pull request #580, base branch `v3`, head `fcfeb791deb014c1612f1d1b336f574a0638a7c3`.

vicondoa:
**Re-verification of head `a12e66062460e1d04874762d71f8e0a93efcfca2`: correct - no findings, merge-ready.**

The fix is confined to the scaffolding function and replaces the fields the base removed with the fields it added, family for family, built with the same test doubles the base own test inputs use. Nothing was deleted: the field list matches the construction inputs at the merge ref exactly, with none missing and none stale, and the factory list carries every converted service, matching the composition helper name for name including the family whose factory is deliberately facet-free.

The validation ordering was verified structurally rather than accepted. At the previous head the branch sat directly on the base tip, so its own tree and the merge ref were the same tree - a pass reported from a divergent tree was impossible - and the literal named fields that did not exist there. At this head the struct is satisfied field for field, so the failure is gone on the tree the pipeline tests. That distinction is the whole point of validating a merge ref rather than a branch in isolation.

Everything previously verified is unchanged in the final tree: both admission zone fences, the top-level identity reads, the removed custody tautology carried into its converted home with the phase gate intact, the credential identity read, and the assertion message that carries the refusal class without an identifier.

---

Review bound to repository `vicondoa/d2b`, pull request #580, base branch `v3`, head `a12e66062460e1d04874762d71f8e0a93efcfca2`.

PR #581, opened 2026-09-21
docs: record the conversion state of every unit in the effect-port programme

vicondoa:
The programme is at the point where the repository should state what it has, rather than what it intended. The plan unit table gains a state table classifying every unit by what is in the tree: converted and merged; converted on a branch and awaiting merge, with the branch named; partially converted with the residual named; retired rather than converted; and outstanding. The task ledger and the debt record carry the same classification, so the three cannot drift apart.

Branch presence never counts as landed, and the distinction the record exists to preserve is between an effect that moved and an effect port that moved while its substantive behaviour stayed in the daemon behind a runtime facet - each partial unit names its residual rather than reading as complete. The known-open structural items are recorded where the next contributor will look: the daemon own Bazel test targets cannot build because rules_rs emits two configurations for one crate, the daemon clippy target fails on otherwise-clean state, and the process-systemd identity read cannot complete on a systemd that no longer serves the main-pid and control-group properties for transient units.


vicondoa:
**Re-verification of head `d28fccfdd8ba33a76004d7fbd1f4d5e32343f1fc`: correct - the record is accurate.**

Both narrow checks pass. The host row carries two facts that are both true at the revision cited and neither overstates nor understates: the family effects trait is defined and implemented only inside its crate, with the crate registration declaring the service and the daemon shared module retaining only the user half; and the daemon-runtime self-checks that form the residual are still present, still exported and still driven from composition at startup, so the row asserts a residual that exists rather than a stale one. It identifies that residual as daemon-runtime self-checks rather than an effect that failed to move, which is what those files are. The framing paragraph, the ledger row and the debt register carry the same two halves in the same order.

The activation row was confirmed by tree state rather than by branch ancestry: the module is absent from the daemon, the crate holds the effects service, and its registration rides the generated table.

One nuance, recorded rather than smoothed: the merge that makes that row true is not in this pull request base, which was cut before it landed. The row is accurate against the mainline tip at the date recorded, and the tree will contain that merge once this lands - stated so a reader comparing against a stale base does not read the row as wrong.

---

Review bound to repository `vicondoa/d2b`, pull request #581, base branch `v3`, head `d28fccfdd8ba33a76004d7fbd1f4d5e32343f1fc`.

PR #582, opened 2026-09-21
test(d2bd): derive the committed startup order from the registration authority

vicondoa:
The test asserting the committed provider startup order failed at the base, because its expected list was a hand-maintained literal that had drifted from the generated registration table. That failure sat in the daemon test surface, which the main suite excludes, so it was invisible to CI while taxing every lane that rebased over it: each had to prove the red test was not theirs.

The registered prefix of the expectation is now built from the same generated table the composition root iterates. A family moving into the table no longer requires a hand edit here, which removes the class of defect rather than this instance of it. The remainder stays literal because it is the historical assembly order in the provider set, which no authority owns.

The derived expectation is built independently of the runtime observed order, so the test still fails if the composed order diverges from the table - reversing the table iteration was used to confirm that before this was written.


vicondoa:
**Independent review of head `c25967a9f13ed8c3e2078219a4acabb788a0c8d7`: correct - no findings.**

The expectation and the observed value come from the same source of truth through different paths, which is what makes the assertion non-vacuous: the expected registered prefix is a compile-time map over the static generated table, while the observed order is produced at runtime by the attach loop appending each provider reference only after a successful attach, with no sorting or deduplication. The reviewer enumerated what would have to change for the test to fail - the composition root ceasing to iterate the table first, the attach loop reordering or skipping, a mangled reference, or any change to the ordered remainder - and confirmed each is caught, including in the literal tail.

One residual is stated rather than glossed: if the generated table were itself generated in the wrong order, the derived prefix would agree with that error, ceding the protection the hand-written literal gave. That is bounded by the table being the declared authority, generated deterministically in a fixed key order behind a drift gate that refuses hand edits, with a parity gate and a startup cross-check that still catches a family silently vanishing. The remaining blind spot is a generator ordering bug, which belongs to the generator.

The improvement is also demonstrated rather than claimed: the base gained a family while this was in review, and the derived expectation absorbed it with no edit - which is exactly the maintenance the hand-written list could not survive.

---

Review bound to repository `vicondoa/d2b`, pull request #582, base branch `v3`, head `c25967a9f13ed8c3e2078219a4acabb788a0c8d7`.

